PR #50, opened 2015-11-05
Allow delete to use a region_alias as well

andrewjhumphrey:
region_aliases are super handy things, I hate remembering  which environment is in which region, so this lets me use them when I delete stacks as well as create them.


gstamp:
:+1:

PR #51, opened 2015-11-05
Handle noecho parameters

gstamp:
Will look something like this.

```
⇒ sm status
Fetching stack information: |==============================================================================================================|
REGION    | STACK_NAME          | STACK_STATUS    | DIFFERENT
----------|---------------------|-----------------|----------
us-east-1 | api-gateway         | UPDATE_COMPLETE | No *
us-east-1 | build-envato-com    | UPDATE_COMPLETE | No
us-east-1 | dewey               | CREATE_COMPLETE | No
us-east-1 | ecs-cluster-stack-b | UPDATE_COMPLETE | Yes
us-east-1 | elasticsearch       | UPDATE_COMPLETE | No *
 * No echo parameters can't be diffed
```

Also removes the puts from the param resolves which was screwing with
the progress bar.

### DIFF
--- a/lib/stack_master/stack_differ.rb
+++ b/lib/stack_master/stack_differ.rb
@@ -18,7 +18,14 @@ module StackMaster
     end
 
     def proposed_parameters
-      YAML.dump(sort_params(@proposed_stack.parameters_with_defaults))
+      # **** out any secret parameters in the current stack.
+      params = @proposed_stack.parameters_with_defaults
+      if @current_stack
+        noecho_keys.each do |key|
+          params[key] = "****"
+        end
+      end
+      YAML.dump(sort_params(params))
     end
 
     def body_different?
@@ -33,10 +40,23 @@ module StackMaster
       if @current_stack
         text_diff('Stack', current_template, proposed_template, context: 7, include_diff_info: true)
         text_diff('Parameters', current_parameters, proposed_parameters)
+        unless noecho_keys.empty?
+          StackMaster.stdout.puts " * can not tell if NoEcho parameters are different, assuming not."
+        end
       else
         text_diff('Stack', '', proposed_template)
         text_diff('Parameters', '', proposed_parameters)
         StackMaster.stdout.puts "No stack found"
+      end
+    end
+
+    def noecho_keys
+      if @current_stack
+        @current_stack.parameters_with_defaults.select do |key, value|
+          value == "****"
+        end.keys
+      else
+        []
       end
     end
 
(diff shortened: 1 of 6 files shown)


lib/stack_master/stack_differ.rb
```diff
@@ -33,6 +40,9 @@ def output_diff
       if @current_stack
         text_diff('Stack', current_template, proposed_template, context: 7, include_diff_info: true)
         text_diff('Parameters', current_parameters, proposed_parameters)
+        unless noecho_keys.empty?
+          StackMaster.stdout.puts " * can not tell if NoEcho parameters are different, assuming not."
```
simpsora: I think we need to assume they _are_ different.

* Worst case if we assume they are different and they're not: stack update happens, stops with "nothing to update".
* Worst case if we assume they are not different and they are: stack not updated with changed params.

lib/stack_master/stack_differ.rb
```diff
@@ -40,6 +50,16 @@ def output_diff
       end
     end
 
+    def noecho_keys
+      if @current_stack
+        @current_stack.parameters_with_defaults.select do |key, value|
```
simpsora: How does this method restrict itself to only params that have `NoEcho` enabled?

gstamp: If AWS has blanked out the value with **** then it's assumed to be a noecho param.  Obviously this will fail someone uses an actual value of **** but that seems unlikely.

lib/stack_master/stack_differ.rb
```diff
@@ -18,7 +18,14 @@ def current_parameters
...
     def proposed_parameters
-      YAML.dump(sort_params(@proposed_stack.parameters_with_defaults))
+      # **** out any secret parameters in the current stack.
+      params = @proposed_stack.parameters_with_defaults
+      if @current_stack
+        noecho_keys.each do |key|
+          params[key] = "****"
+        end
+      end
+      YAML.dump(sort_params(params))
```
stevehodgkiss: This will affect the apply command too. Is that intentional?

gstamp: Just the diff of the apply command which I think should be okay.

lib/stack_master/stack_differ.rb
simpsora: Ah, right.  `select`.  Carry on.

stevehodgkiss:
:+1:

PR #54, opened 2015-11-12
Ensure StackMaster uses the new ruby AWS SDK

stevehodgkiss:
@andrewjhumphrey reported an issue where an error is thrown about a missing `Aws` constant. It's most likely because we're not specifying the aws-sdk version and it's using an old ruby sdk.


andrewjhumphrey:
+1 if we depend on a version of a library we should call it out :-)

PR #55, opened 2015-11-12
commands get passed an options array

andrewjhumphrey:
also update spec to reflect this for validate, which was failing when called
``` MacBook-Pro:stackmaster andrewhumphrey$ sm validate staging 
marketdbreplicas --trace Received the following args: 3 0:StackMaster::Config 
1:StackMaster::StackDefinition 2:
/Users/andrewhumphrey/src/envato-cloud-management/.bundle/ruby/2.1.0/gems/stack_master-0.0.2/lib/stack_master/commands/validate.rb:7:in
`initialize': wrong number of arguments (3 for 2) (ArgumentError)
```

### DIFF
--- a/lib/stack_master/commands/validate.rb
+++ b/lib/stack_master/commands/validate.rb
@@ -4,7 +4,7 @@ module StackMaster
       include Command
       include Commander::UI
 
-      def initialize(config, stack_definition)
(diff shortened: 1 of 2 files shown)


stevehodgkiss:
:+1:

PR #56, opened 2015-11-12
Improve error messages when parameters are broken

andrewjhumphrey:
Before:
``` MacBook-Pro:stackmaster andrewhumphrey$ sm apply staging 
market-db-replicas error: 5. Use --trace to view backtrace
```

After:
``` MacBook-Pro:stackmaster andrewhumphrey$ sm apply staging 
market-db-replicas Unable to resolve parameter "DbAllocatedStorage" value causing error: 5
```

### DIFF
--- a/lib/stack_master/parameter_resolver.rb
+++ b/lib/stack_master/parameter_resolver.rb
@@ -19,8 +19,8 @@ module StackMaster
         begin
           parameters[key] = resolve_parameter_value(value)
         rescue InvalidParameter
-          say "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
-          exit 1
+          StackMaster.stderr.puts  "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
+          raise
         end
         parameters
       end
--- a/lib/stack_master/parameter_resolver.rb
+++ b/lib/stack_master/parameter_resolver.rb
@@ -16,7 +16,12 @@ module StackMaster
 
     def resolve
       @parameters.reduce({}) do |parameters, (key, value)|
-        parameters[key] = resolve_parameter_value(value)
+        begin
+          parameters[key] = resolve_parameter_value(value)
+        rescue InvalidParameter
+          say "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
+          exit 1
+        end
         parameters
       end
     end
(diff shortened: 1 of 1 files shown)


lib/stack_master/parameter_resolver.rb
```diff
@@ -16,7 +16,12 @@ def initialize(config, stack_definition, parameters)
 
     def resolve
       @parameters.reduce({}) do |parameters, (key, value)|
-        parameters[key] = resolve_parameter_value(value)
+        begin
+          parameters[key] = resolve_parameter_value(value)
+        rescue InvalidParameter
+          say "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
```
gstamp: I don't think say will work here.

gstamp: You probably want

```
StackMaster.stdout.puts 
```

andrewjhumphrey: Say works, but I suspect I want to do that using stderr instead of stdout.

gstamp: Actually I'm thinking it would be better to raise an error with a good
error message and let it bubble to the top.

andrewjhumphrey:
Output is now:
```
MacBook-Pro:stackmaster andrewhumphrey$ sm apply staging market-db-replicas 
Unable to resolve parameter "DbAllocatedStorage" value causing error: 5
error: 5. Use --trace to view backtrace
```

gstamp:
Sorry to be painful but I think changing this:

``` ruby
StackMaster.stderr.puts  "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
raise
```

to this

``` ruby
raise "Unable to resolve parameter #{key.inspect} value causing error: #{$!.message}"
```

Since the message will get printed for you anyway.

andrewjhumphrey:
I didn't do that because it breaks the specs,  if it's worth it I can fix the specs:
```
  1) StackMaster::ParameterResolver it throws an error when the hash contains more than one key
     Failure/Error: expect {
       expected StackMaster::ParameterResolver::InvalidParameter, got #<RuntimeError: Unable to resolve parameter :param value causing error: {:nested1=>"value1", :nested2=>"value2"}> with backtrace:
         # ./lib/stack_master/parameter_resolver.rb:22:in `rescue in block in resolve'
```

andrewjhumphrey:
Ahh, I can fix that if I type the raise.

andrewjhumphrey:
And the output is how we actually want it to look:
MacBook-Pro:stackmaster andrewhumphrey$ sm apply staging market-db-replicas 
error: Unable to resolve parameter "DbAllocatedStorage" value causing error: 5. Use --trace to view backtrace

gstamp:
Great!

PR #57, opened 2015-11-19
Increase the maximum allowed width of output

gstamp:
Output variables tend to get cut off with the default.

/cc @stevehodgkiss


stevehodgkiss:
:+1:

PR #58, opened 2015-12-14
Dynamically require parameter resolvers

lethalpaga:
This allows the creation of new resolvers in a separate gem (See the README for more details).

My use case is that we have an internal company-specific service to lookup Vpc and subnet Ids

### DIFF
--- a/lib/stack_master/parameter_resolver.rb
+++ b/lib/stack_master/parameter_resolver.rb
@@ -27,19 +27,53 @@ module StackMaster
 
     private
 
+    def require_parameter_resolver(file_name)
+      begin
+        require "stack_master/parameter_resolvers/#{file_name}"
+      rescue LoadError
+        raise ResolverNotFound.new(file_name)
+      end
+    end
+
+    def load_parameter_resolver(class_name)
+      begin
+        # Check if the class name already exists
+        return if resolver_class_const(class_name)
+      rescue NameError
+        # If it doesn't, try to load it
+        require_parameter_resolver(class_name.underscore)
+      end
+    end
+
     def resolve_parameter_value(parameter_value)
       return parameter_value if String === parameter_value || parameter_value.nil?
       raise InvalidParameter, parameter_value unless Hash === parameter_value
       raise InvalidParameter, parameter_value unless parameter_value.keys.size == 1
-      resolver_class_name = parameter_value.keys.first.to_s.camelize
+
+      resolver_name = parameter_value.keys.first.to_s
+      load_parameter_resolver(resolver_name)
+
       value = parameter_value.values.first
-      resolver_class(resolver_class_name).resolve(value)
+      resolver_class_name = resolver_name.camelize
+      call_resolver(resolver_class_name, value)
+    end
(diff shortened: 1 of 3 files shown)


spec/stack_master/parameter_resolver_spec.rb
```diff
@@ -55,4 +63,29 @@ def resolve(params)
       expect(resolve(param: { my_resolver: 2 }, param2: { my_resolver: 2 })).to eq(param: 10, param2: 10)
     end
   end
+
+  context 'when the resolver class already exist' do
+    it 'does not try to load it' do
+      expect(subject).to receive(:load_parameter_resolver).once.and_call_original
+      expect(subject).not_to receive(:require_parameter_resolver)
+
+      subject.resolve
```
stevehodgkiss: This would read better with named subjects - `subject(:parameter_resolver) { .. }`

lib/stack_master/parameter_resolver.rb
```diff
@@ -27,19 +27,53 @@ def resolve
...
+    end
+
+    def load_parameter_resolver(class_name)
+      begin
+        # Check if the class name already exists
+        return if resolver_class_const(class_name)
+      rescue NameError
+        # If it doesn't, try to load it
+        require_parameter_resolver(class_name.underscore)
+      end
```
stevehodgkiss: The explicit `begin/end` lines aren't necessary here.

lib/stack_master/parameter_resolver.rb
```diff
@@ -27,19 +27,53 @@ def resolve
...
+        require "stack_master/parameter_resolvers/#{file_name}"
+      rescue LoadError
+        raise ResolverNotFound.new(file_name)
+      end
+    end
+
+    def load_parameter_resolver(class_name)
+      begin
+        # Check if the class name already exists
+        return if resolver_class_const(class_name)
```
stevehodgkiss: Any reason for the conditional return? I think it would work without it, the difference being the return value set to the constant instead of nil.

lib/stack_master/parameter_resolver.rb
```diff
@@ -27,19 +27,53 @@ def resolve
 
     private
 
+    def require_parameter_resolver(file_name)
+      begin
+        require "stack_master/parameter_resolvers/#{file_name}"
+      rescue LoadError
+        raise ResolverNotFound.new(file_name)
+      end
```
stevehodgkiss: `begin/end` lines aren't necessary here.

stevehodgkiss:
Nitpicks aside - this is great, thanks for the contribution!

lib/stack_master/parameter_resolver.rb
lethalpaga: Not really, it's a leftover from my previous implementation. I'll remove it

lethalpaga:
Everything should be fixed :hammer:

gstamp:
:+1:

stevehodgkiss:
Released 0.0.4 with these changes.

PR #59, opened 2015-12-14
Add Travis support

lethalpaga:
* Add rake tasks to run rspec and cucumber tests (theses tasks are not run by default to keep the same behaviour)
 * Add a configuration file for [Travis](https://travis-ci.org)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: ruby
+rvm:
+  - 2.2
+sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ language: ruby
 rvm:
   - 2.2
 sudo: false
+script:
+  - bundle exec rake spec features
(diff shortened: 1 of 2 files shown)


lethalpaga:
To elaborate a bit more on my motivations for doing this :

 * I've seen that you are currently using buildkite, unfortunately it's not free, making it harder for the community to contribute to this project.
 * This change isn't intended to replace Buildkite with Travis but to allow contributors to run the tests on their own forks.
 * The reason I've added the rake tasks is that Travis uses a simple yaml file to describe the build using a profile (here 'ruby') and automatically takes care of the bundle installs, runs tests on multiple versions of ruby, etc.
    As it's probably not a good idea to have a different paths for each CI system, I think it would be better to change the buildkite scripts to run rake rather than to twist the logic of travis and do a custom build. I didn't do it in this PR as I'd first like to know what you think about it.

I'm happy to discuss this further if you want.

gstamp:
The reason we used buildkite was simply because at the time we hadn't yet open sourced the code. Now that it's open source there's no good reason to stick with it.

:+1:

stevehodgkiss:
:+1: let's move to travis.

PR #61, opened 2015-12-22
Allow arrays of security groups to be resolved

lethalpaga:
Some resources (load balancers, Ec2 instances, ...) take an array of security groups as a parameter. This is an extension of the security_group resolver to do that.

This is will return a single value if a single value is provided, and an array if an array is provided, so it is backward-compatible


stevehodgkiss:
Hi @lethalpaga 

I think it would be better to create a new resolver called `security_groups` instead of overloading this one. That resolver could just use the singular `security_group` resolver to resolve a list of security groups.

This would be really useful to build an of stack outputs, for example this as input to a `List<AWS::EC2::Subnet::Id>` parameter:

```
public_subnets:
  stack_outputs:
    - my-vpc/public_subnet_1
    - my-vpc/public_subnet_2
```

It might be worth noting that this will only work with the `List<...>` parameter types (AFAIK, at least).

Thanks for the PR!

lethalpaga:
Hi,
I've finally found some time to update this PR :
 * Created a generic resolver 'arrayizer'
 * Created a base class for resolvers. This adds an `array_resolver` method that can be used in the class declaration to generate the corresponding array resolver (so this would generate resolvers for security_group**s**, stack_output**s**, etc.)
 * That method is used on all existing resolvers (because why not?)

gstamp:
:+1: Neat solution.

Unfortunately @stevehodgkiss is away at the moment but I think this looks good to merge so I'll probably hit the button on this soon.

lethalpaga:
Thanks for merging!

Now that I think about it, I should have put that array_resolver method in a module and extended the resolver classes with it instead of inheriting. I'm happy to do a new PR with that if you want to.

stevehodgkiss:
Happy for you to do a new PR with further refactorings @lethalpaga. Thanks for your contributions.

PR #62, opened 2016-01-19
Fix stack deletion when using a region alias

lethalpaga:
Region aliases don't work for stack deletion if the config file isn't explicitly specified with the -c option.
```
bundle exec stack_master delete my_region_alias my_stack
error: getaddrinfo: Name or service not known. Use --trace to view backtrace
```
Stack_master passes the literal 'my_region_alias' string to the AWS SDK. This is due to the following line ;
```patch
diff --git a/lib/stack_master/cli.rb b/lib/stack_master/cli.rb
index 0e45c3c..539c4ce 100644
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -130,17 +131,20 @@ module StackMaster
-         # Because delete can work without a stack_master.yml
-         if options.config and File.file?(options.config)
```
options.config in nil and the call to `load_config` that falls back to `stack_master.yml` if no config file is provided is called after that check. 

I solved this by providing a default value to options.config for every action (to be consistent).

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -12,6 +12,10 @@ module StackMaster
       TablePrint::Config.io = StackMaster.stdout
     end
 
+    def default_config_file
+      "stack_master.yml"
+    end
+
     def execute!
       program :name, 'StackMaster'
       program :version, '0.0.1'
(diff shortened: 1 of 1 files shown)


gstamp:
:+1:

Issue #60, opened 2015-12-20
Feature request - Nested stacks creation


lethalpaga:
At the moment stack_master cannot create nested stacks as it's supplying the template body to the API.
Do you have any plans to support that?

The way I see it :
- An S3 bucket URL and prefix would have to be added to the config file (global and per-region configuration)
- All nested templates would have to be declared alongside the main template, maybe as a `file` parameter (that would allow one to upload any kind of file to that bucket, if used by a UserData script or such) and uploaded to that bucket
- When creating the stack, pass in a `template_url` instead of a `template_body` parameter
- I'm not sure how the stack events would be handled. Worst case would be that only the events from the toplevel stack would be displayed (so not events from the nested stacks). I suppose it's good enough for a first version
- Stack deletion should work as-is as cfn takes care of the substack deletion
- Diffs, Output and Resources listing would only show the ones from the toplevel template. AFAIK there is no easy way to get substacks other that iterating over the Cloudformation::Stack resources recursively. Again I assume that would be acceptable in a first version.
- The `too_big?`method would need to know which method is used as the allowed size is different

Any thoughts on this? If not I'm happy to give a crack at this.

stevehodgkiss:
Hi @lethalpaga 

It sounds like there are two parts to this

- allowing CF updates via S3
- nested stack support

We don't have any immediate plans to implement either, but we've discussed CF updates through S3 before. We don't use nested stacks and instead connect stacks together with stack outputs/parameter resolvers.

> All nested templates would have to be declared alongside the main template, maybe as a file parameter (that would allow one to upload any kind of file to that bucket, if used by a UserData script or such) and uploaded to that bucket

That sounds reasonable. Just thinking out load: I wonder if we could use parameter resolvers for this. For example:

```
nested_stack_url:
  nested_stack: template-name.rb
```

The resolver would automatically compile and upload the template and provide an S3 url. I'm not sure how diffing would work on nested stacks though.

Happy for you to have a go implementing this. Thanks!

luckymike:
The SparkleFormation `nest!()` method will generate a nested stack resource (which includes a path to the template in the S3 Bucket). You would need to support configuring the bucket(s) to use (and providing these to SparkleFormation at compilation) and uploading templates to the bucket. 

This is implemented in [sfn](https://github.com/sparkleformation/sfn) if you'd like to see it in action to understand how the parts work.

patrobinson:
[...]

PR #63, opened 2016-01-21
Add the capability to use templates from S3

lethalpaga:
This covers the first part of #60, which is the upload to S3 of templates and associated files, and triggering the stack creation from an S3 URL.

 *This works for manually created nested templates, I haven't tested the SparkleFormation nested stacks functions. As @luckymike mentioned the s3 bucket/prefix would have to be passed on to sfn.
 * Stack diff and events tail only apply to the toplevel stack for now (when using nested stacks).
 * If no bucket is configured the old behaviour applies

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -78,7 +78,7 @@ module StackMaster
       end
 
       def stack_too_big
-        if proposed_stack.too_big?
+        if proposed_stack.too_big?(use_s3?)
           StackMaster.stdout.puts 'The (space compressed) stack is larger than the limit set by AWS. See http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html'
           true
         else
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -7,7 +7,7 @@ module StackMaster
 
       def initialize(config, stack_definition, options = {})
         @config = config
-        @s3_config = stack_definition['s3']
+        @s3_config = stack_definition.s3
         @stack_definition = stack_definition
         @from_time = Time.now
         @updating = false
@@ -52,7 +52,7 @@ module StackMaster
       end
 
       def use_s3?
-        @s3_config
+        !@s3_config.empty?
       end
 
       def diff_stacks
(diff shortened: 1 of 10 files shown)


lib/stack_master/config.rb
```diff
@@ -74,7 +74,7 @@ def load_stacks(stacks)
         region = Utils.underscore_to_hyphen(region)
         stacks_for_region.each do |stack_name, attributes|
           stack_name = Utils.underscore_to_hyphen(stack_name)
-          stack_attributes = build_stack_defaults(region).deeper_merge(attributes).merge(
+          stack_attributes = build_stack_defaults(region).deeper_merge!(attributes).merge(
```
lethalpaga: I had to use depper_merge! to allow redefining

stevehodgkiss: I'm not sure I understand. Are you saying that the default s3 config couldn't be overridden on a per-stack basis or that the override bled into other stack definitions?

lib/stack_master/commands/apply.rb
```diff
@@ -85,14 +95,43 @@ def create_stack
...
+      end
+
+      def template_value
+        return proposed_stack.maybe_compressed_template_body unless use_s3?
+        s3.url(@s3_config.merge('template' => @stack_definition.template))
+      end
+
+      def files_to_upload
+        return [] unless use_s3?
+        [@stack_definition.files_to_upload, @stack_definition.template_file_path].flatten
```
stevehodgkiss: We wouldn't want to upload the raw `@stack_definition.template_file_path` for sparkleformation templates. I think we should use `proposed_stack.maybe_compressed_template_body` instead.

stevehodgkiss:
Looks good! It'd be great to have some cucumber coverage of this new behaviour :)

lib/stack_master/config.rb
lethalpaga: The default s3 config couldn't be overridden on a per-stack basis. According to the [deep_merge documentation](https://github.com/danielsdeleo/deep_merge/blob/master/lib/deep_merge/deep_merge_hash.rb#L18) this is the expected result when using `deep_merge` (as opposed to `deep_merge!`).

stevehodgkiss: Cool, we definitely want this change then :)

stevehodgkiss:
Hey @lethalpaga, will you have time to pick this up again at some point? If not I'd be happy to pick it up.

lethalpaga:
I'm not sure when I'll have time to work on this. Feel free to pick it up if you want.

PR #64, opened 2016-01-27
adjust the Prompter to use #gets instead of #getch, so we can pipe the 'yes' command to stack_master and run autonomously

redterror:
Also slightly expand the acceptable affirmative answers to include "yes" and ignore case.

### DIFF
--- a/lib/stack_master/prompter.rb
+++ b/lib/stack_master/prompter.rb
@@ -5,10 +5,9 @@ module StackMaster
       answer = if ENV['STUB_AWS']
         ENV['ANSWER']
       else
-        STDIN.getch.chomp
+        STDIN.gets.chomp
       end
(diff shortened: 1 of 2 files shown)


stevehodgkiss:
Would a new command line option `-y` (like `apt-get`) make more sense for this scenario?

redterror:
I considered that, but this was almost zero work and did the job.

lethalpaga:
Funnily enough I was just working on the same issue. My solution is still a WIP and I don't know when I'll have time to work on it.
FYI https://github.com/envato/stack_master/compare/master...lethalpaga:fix/interactive
Feel free to vampirise it if you want.

stevehodgkiss:
Closing in favour of https://github.com/envato/stack_master/pull/71

Thanks!

PR #65, opened 2016-01-28
use ENV['COLUMNS'] or 'tput cols' to detect the terminal width.  fall back to 80

redterror:
I wasn't sure the best way to write a spec for this, so feedback is welcome there.  My CF setup has very long stack names so the 80 column default is painful, hence this patch.


redterror:
Would other commands benefit as well?

lib/stack_master/commands/helper.rb
```diff
@@ -0,0 +1,21 @@
+require 'active_support/concern'
+
+module StackMaster
+  module Commands
+    module Helper
+      extend ActiveSupport::Concern
+
+      included do
```
stevehodgkiss: I'm not sure the use of active support concern is necessary here. Could we just use regular ruby to avoid depending on AS?

```
module Helper
  def window_size
    ...
  end
end
```

lib/stack_master/commands/helper.rb
```diff
@@ -0,0 +1,21 @@
+require 'active_support/concern'
+
+module StackMaster
+  module Commands
+    module Helper
```
stevehodgkiss: `Helper` is a pretty generic name. Perhaps rename to `TerminalHelper` or something more descriptive?

lib/stack_master/commands/helper.rb
```diff
@@ -0,0 +1,21 @@
...
+      extend ActiveSupport::Concern
+
+      included do
+        def window_size
+          size = ENV["COLUMNS"] || `tput cols`.chomp
```
stevehodgkiss: Nitpick: I would prefer to use `ENV.fetch('COLUMNS') { .. }` for this behaviour.

lib/stack_master/commands/helper.rb
redterror: You've already got `active_support` in the Gemfile, I figured that made it fair game.  :-P

lib/stack_master/commands/helper.rb
redterror: Neat, I didn't know that construct!

redterror:
I pushed changes incorporating your comments, but I think the `ActiveSupport::Concern` pattern is a better one than just using the module directly, especially given that the gem is already there.

Issue #66, opened 2016-01-28
Numeric parameters not handled well


gstamp:
When you pass a numeric parameter to Stack Master it blows up with something like this:

```
src/stack_master/lib/stack_master/parameter_resolver.rb:22:in `rescue in block in resolve': Unable to resolve parameter "AutoScalingGroupMinSize" value causing error: 2 (StackMaster::ParameterResolver::InvalidParameter)
```

A work around is to just convert those parameters to strings but it would be nice not to have to worry about this.

stevehodgkiss:
This is fixed now.

PR #67, opened 2016-01-28
Allow stack master to handle numeric parameters

gstamp:
Fixes: https://github.com/envato/stack_master/issues/66

Although you can specify numeric parameters in stacks AWS when you pass those parameters they need to be strings. Passing numeric parameters to SM causes a confusing error. This PR fixes this problem by letting you pass numeric params then converting them to strings for AWS.

/cc @stevehodgkiss

### DIFF
--- a/lib/stack_master/parameter_resolver.rb
+++ b/lib/stack_master/parameter_resolver.rb
@@ -42,7 +42,7 @@ module StackMaster
     end
 
     def resolve_parameter_value(parameter_value)
-      return parameter_value if String === parameter_value || parameter_value.nil?
+      return parameter_value if Numeric === parameter_value || String === parameter_value || parameter_value.nil?
       raise InvalidParameter, parameter_value unless Hash === parameter_value
       raise InvalidParameter, parameter_value unless parameter_value.keys.size == 1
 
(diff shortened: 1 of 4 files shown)


nogates:
Nice! :+1:

lib/stack_master/utils.rb
```diff
@@ -4,7 +4,7 @@ module Utils
 
     def hash_to_aws_parameters(params)
       params.inject([]) do |params, (key, value)|
-        params << { parameter_key: key, parameter_value: value }
+        params << { parameter_key: key, parameter_value: value.to_s }
```
stevehodgkiss: I think this will break arrays (parameter resolvers that use ResolverArray).

lib/stack_master/parameter_resolver.rb
```diff
@@ -42,7 +42,7 @@ def load_parameter_resolver(class_name)
     end
 
     def resolve_parameter_value(parameter_value)
-      return parameter_value if String === parameter_value || parameter_value.nil?
+      return parameter_value if Numeric === parameter_value || String === parameter_value || parameter_value.nil?
```
stevehodgkiss: Could we move this to a separate line with something like this? `return parameter_value.to_s if Numeric === parameter_value`

lib/stack_master/utils.rb
gstamp: Thanks. Will look into that. I guess we need some extra test coverage around that case.

stevehodgkiss:
Closing because I've addressed the issue mentioned in this PR and added support for numeric types here -> https://github.com/envato/stack_master/pull/70

PR #69, opened 2016-02-03
handle throttling errors from AWS with a simple capped backoff strategy

redterror:
I have a CF setup with a large number of stacks, and sometimes AWS throws errors b/c the request rate on the `status` command is too high.  This fixes that by adding a very simple capped-delay retry strategy and related tests.

### DIFF
--- a/lib/stack_master/aws_driver/cloud_formation.rb
+++ b/lib/stack_master/aws_driver/cloud_formation.rb
@@ -11,7 +11,9 @@ module StackMaster
       end
 
       def describe_stacks(options)
-        cf.describe_stacks(options)
+        retry_with_backoff do
+          cf.describe_stacks(options)
+        end
       end
 
       def cancel_update_stack(options)
@@ -50,6 +52,23 @@ module StackMaster
 
       def cf
         @cf ||= Aws::CloudFormation::Client.new(region: @region)
+      end
+
+      def retry_with_backoff
+        delay       = 1
+        max_delay   = 30
+        begin
+          yield
+        rescue Aws::CloudFormation::Errors::Throttling => e
+          if e.message =~ /Rate exceeded/
+            sleep delay
+            delay *= 2
+            if delay > max_delay
+              delay = max_delay
+            end
+            retry
+          end
+        end
       end
     end
   end
--- a/spec/stack_master/commands/status_spec.rb
+++ b/spec/stack_master/commands/status_spec.rb
@@ -8,11 +8,14 @@ RSpec.describe StackMaster::Commands::Status do
 
   before do
     allow(Aws::CloudFormation::Client).to receive(:new).with(region: 'us-east-1').and_return cf
-    allow(StackMaster::Stack).to receive(:find).and_return stack1, stack2
-    allow(StackMaster::Stack).to receive(:generate).and_return proposed_stack1, proposed_stack2
   end
 
   context "#perform" do
+    before do
+      allow(StackMaster::Stack).to receive(:find).and_return stack1, stack2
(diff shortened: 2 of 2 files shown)


lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -51,6 +53,23 @@ def validate_template(options)
...
       end
+
+      def retry_with_backoff
+        delay       = 1
+        max_delay   = 30
+        begin
+          yield
+        rescue Aws::CloudFormation::Errors::Throttling => e
+          if e.message =~ /Rate exceeded/
+            sleep delay
```
stevehodgkiss: Is it worth outputting that API requests are being throttled?

stevehodgkiss:
Thanks! :+1:

lib/stack_master/aws_driver/cloud_formation.rb
redterror: Ehhh, I see both sides of it.  Here are my thoughts:

Pro:
* I like transparency
* Blames AWS for slowness, not stackmaster (yay blame!)

Con:
* Not actionable info to the user in any way (currently).
* Screen clutter

In my particular use case we have quite a lot of stacks, so this happens frequently, and beyond the first time it happens the novelty is lost.  Hence I opted for silence.  But I don't feel strongly about it, so as always, feedback is welcome.

stevehodgkiss: Makes sense, let's go with the silent approach for now then.

lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -11,7 +11,9 @@ def delete_stack(options)
       end
 
       def describe_stacks(options)
-        cf.describe_stacks(options)
+        retry_with_backoff do
+          cf.describe_stacks(options)
+        end
       end
```
stevehodgkiss: Should we wrap all AWS API requests with this retry logic?

redterror: I think not yet.  At least in my experience / usage pattern so far, I haven't had that problem outside of `status` and adding a fix w/o a demonstrated problem is something I like to avoid.

That said, backing up slightly, my workflow / use-case so far has been to do something vaguely like:

```
$ stack_master status | grep -v _COMPLETE | awk '{print $3}' > foo
$ vi foo # trim to taste
$ for i in `cat foo` ; do yes | stack_master apply my-region $i ; done
```

This lets me load a pile of stacks in an unattended fashion.  This does require my `feat/use_gets_instead_of_getch` PR so the `yes` pipe works.

If `apply` accepted a list of stacks, or could work out dependencies among stacks, I think then it would be necessary/useful.  In its current state I don't think its strictly required.  My team and I have discussed building that functionality.

stevehodgkiss: So that's to apply any dirty stack? Perhaps that could be a new feature/cmd line option.

```
stack_master -f apply --changed
stack_master -f diff --changed
```

I guess that would require two separate bits of work. One to allow `stack_master apply` to work without parameters and simply loop through stacks applying them. Then adding `--changed` to only operate on stacks if the stack is different to what is on cloudformation.

`-f` will be introduced in https://github.com/envato/stack_master/pull/71

Working out dependencies would require either manual declaration of dependencies in stack_master.yml or working it out by looking at parameter files and stack_output usage. That would be a welcome addition though :)

redterror: > Working out dependencies would require either manual declaration of dependencies in stack_master.yml or working it out by looking at parameter files and stack_output usage. That would be a welcome addition though :)

Awesome, we're on the same page then in terms of our approach!

PR #70, opened 2016-02-06
Allow arrays in parameter files and support numeric types

stevehodgkiss:
Arrays can be supplied to the AWS SDK for the List<> types, so StackMaster should support that.

Also allows numeric types to be passed in and converted to strings before being passed to the AWS SDK . Ref: https://github.com/envato/stack_master/pull/67 and https://github.com/envato/stack_master/issues/66


gstamp:
:+1:

PR #71, opened 2016-02-06
Add `-y` and `-n` global options to run without user interaction

stevehodgkiss:
Accepts `-y` to force a 'y' response to any prompt in StackMaster and cleans up features.

Fixes https://github.com/envato/stack_master/pull/64

Took inspiration from @lethalpaga's work on this here https://github.com/envato/stack_master/compare/master...lethalpaga:fix/interactive

cc @redterror

### DIFF
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -66,6 +66,9 @@ module StackMaster
     @non_interactive = true
   end
 
+  attr_accessor :non_interactive_answer
+  @non_interactive_answer = 'y'
+
   def base_dir
     File.expand_path(File.join(File.dirname(__FILE__), ".."))
   end
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -86,6 +86,10 @@ module StackMaster
     @stdout = io
   end
 
+  def stdin
+    $stdin
+  end
+
   def stderr
     @stderr || $stderr
   end
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -54,6 +54,18 @@ require "stack_master/cli"
 module StackMaster
   extend self
 
+  def interactive?
+    !non_interactive?
+  end
+
+  def non_interactive?
+    @non_interactive || false
+  end
+
+  def non_interactive!
+    @non_interactive = true
+  end
+
   def base_dir
     File.expand_path(File.join(File.dirname(__FILE__), ".."))
   end
--- a/lib/stack_master/prompter.rb
+++ b/lib/stack_master/prompter.rb
@@ -3,7 +3,12 @@ module StackMaster
     def ask?(question)
       StackMaster.stdout.print question
       answer = if StackMaster.interactive?
-        STDIN.getch.chomp
+        if StackMaster.stdin.tty?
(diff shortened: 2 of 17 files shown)


lib/stack_master.rb
```diff
@@ -54,6 +54,18 @@
 module StackMaster
   extend self
 
+  def interactive?
+    !non_interactive?
```
redterror: It may be worthwhile to attempt to check if stdout is a tty.  So:

`!non_interactive? && $stdout.tty?`

stevehodgkiss: I'm not sure if defaulting to yes in that case would be expected behaviour though. Perhaps we could handle that case and output a message about the `-f` option.

stevehodgkiss: Added  6b8d6f1 to default to no and add a message about `-f`.

stevehodgkiss:
Ping @gstamp

lib/stack_master/prompter.rb
```diff
@@ -1,11 +1,17 @@
...
+        if StackMaster.stdin.tty? && StackMaster.stdout.tty?
+          StackMaster.stdin.getch.chomp
+        else
+          StackMaster.stdout.puts
+          StackMaster.stdout.puts "STDOUT or STDIN was not a TTY. Defaulting to no. To force yes use -f"
+          'n'
+        end
       else
-        STDIN.getch.chomp
+        ENV.fetch('ANSWER') { 'y' }
```
nemski: Why not use another command line option to answer questions, rather than an environment variable?

stevehodgkiss: The default without `-f` is no, with `-f` yes. ANSWER is used by our test suite to fake an answer. It might be cleaner if we introduced `StackMaster.default_non_interactive_answer` and simply stubbed it in integration tests instead.

nemski: Or you could provide two command line options, `y` and `n`. I think changing behaviour based on an ENV variable as generic as `ANSWER` might cause unexpected behaviour (since it's also not really documented). If you do use an ENV variable I'd prefer something like `SM_ANSWER`

stevehodgkiss: Sounds good. I've implemented your suggestion.

lib/stack_master/prompter.rb
```diff
@@ -1,11 +1,17 @@
 module StackMaster
   module Prompter
     def ask?(question)
-      StackMaster.stdout.print question
-      answer = if ENV['STUB_AWS']
-        ENV['ANSWER']
+      answer = if StackMaster.interactive?
+        StackMaster.stdout.print question
+        if StackMaster.stdin.tty? && StackMaster.stdout.tty?
```
gstamp: Is it necessary to check both stdout and in?

gstamp:
:+1:

nemski:
:+1:

lib/stack_master/prompter.rb
stevehodgkiss: I think so. Otherwise it would fail in scenario's like this:

```
stack_master apply us-east-1 my-stack > out.txt
```

PR #74, opened 2016-02-07
Allow commands to run on a single stack, multiple or all stacks

stevehodgkiss:
Allows running commands on multiple or all stacks. Examples:

```
stack_master diff # diffs all stacks
stack_master apply # applies all stacks
stack_master apply us-east-1 # applies all stacks in us-east-1
stack_master apply us-east-1 my-stack us-east-1 my-other stack # applies the two stacks. I need to add a cucumber scenario for this.
```

Next up will be adding a `--changed` option to only execute commands on changed stacks. This will be useful to run on CI to post stack diffs to PRs. `stack_master diff --changed`.

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -33,7 +33,7 @@ module StackMaster
         c.example 'update a stack named myapp-vpc in us-east-1', 'stack_master apply us-east-1 myapp-vpc'
         c.action do |args, options|
           options.defaults config: default_config_file
-          execute_stack_command(StackMaster::Commands::Apply, args, options)
+          execute_stacks_command(StackMaster::Commands::Apply, args, options)
         end
       end
 
@@ -43,7 +43,7 @@ module StackMaster
         c.description = "Displays outputs for a stack"
         c.action do |args, options|
           options.defaults config: default_config_file
-          execute_stack_command(StackMaster::Commands::Outputs, args, options)
+          execute_stacks_command(StackMaster::Commands::Outputs, args, options)
         end
       end
 
@@ -69,7 +69,7 @@ module StackMaster
         c.example 'diff a stack named myapp-vpc in us-east-1', 'stack_master diff us-east-1 myapp-vpc'
         c.action do |args, options|
           options.defaults config: default_config_file
-          execute_stack_command(StackMaster::Commands::Diff, args, options)
+          execute_stacks_command(StackMaster::Commands::Diff, args, options)
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/stack_master/cli.rb
```diff
@@ -174,22 +174,21 @@ def load_config(file)
...
-      aliased_region, stack_name = args
-      region = Utils.underscore_to_hyphen(config.unalias_region(aliased_region))
-      stack_name = Utils.underscore_to_hyphen(stack_name)
-      StackMaster.cloud_formation_driver.set_region(region)
-      stack_definition ||= config.find_stack(region, stack_name)
-      if stack_definition.nil?
-        say "Could not find stack definition #{stack_name} in region #{region}"
-        return
+      args = [nil, nil] if args.size == 0
+      args.each_slice(2) do |aliased_region, stack_name|
```
andrewjhumphrey: Neat way to work on the pairs of arguments!   
I'd also like to be able to do something like:
```stack_master apply us-east-1  stack-1 stack-2 stack-3```
But that's just syntatic sugar I guess :-)

**Comment by andrewjhumphrey** [comment]
+1 no matter which way the question I posed is answered.

**Comment by gstamp** [comment]
I'm unclear about the code path for:

`stack_master apply us-east-1`

Is that just getting destructured to region, nil? If so... nice!

We should probably add something to the readme about this?

:+1:

File: lib/stack_master/cli.rb (hunk 1 above)
Comment (gstamp) [inline lib/stack_master/cli.rb]: Could be a bit ambiguous with the the region names vs stack names.

Maybe something like

`stack_master apply us-east-1 stack-1,stack-2,stack-3`

Comment (stevehodgkiss) [inline lib/stack_master/cli.rb]: Hmmm. That might be tricky or come with unexpected behaviour. Currently this PR enables:

```
stack_master apply us-east-1 stack-1 us-east-1 stack-2 us-east-1 stack-3
```

**Comment by stevehodgkiss** [comment]
@gstamp yep:

```
❯ irb
irb(main):001:0> ['us-east-1'].each_slice(2).to_a
=> [["us-east-1"]]

irb(main):002:0> ['us-east-1'].each_slice(2) {|r,s| puts "r: #{r}, s: #{s.class}" }
r: us-east-1, s: NilClass
=> nil
irb(main):003:0>
```

Good point I'll update the readme before merging.

lib/stack_master/cli.rb
stevehodgkiss: @gstamp that solution sounds reasonable. I'll leave it for another PR though.

nemski:
:+1:

PR #75, opened 2016-02-07
Allow list to output in a parsable format

nemski:
So that a script can output a list of stacks and operate on each one

### DIFF
--- a/features/list.feature
+++ b/features/list.feature
@@ -1,0 +1,34 @@
+Feature: List command
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us_east_1:
+          stack1:
+            template: stack1.json
+          stack2:
+            template: stack2.json
+          stack3:
+            template: stack3.json
+      """
+
+  Scenario: Run list command and get a list of stacks
+    When I run `stack_master list` interactively
+    Then the output should contain all of these lines:
+      | REGION    | STACK_NAME|
+      | ----------|---------- |
+      | us-east-1 | stack1    |
+      | us-east-1 | stack2    |
+      | us-east-1 | stack3    |
+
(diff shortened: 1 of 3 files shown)


stevehodgkiss:
:+1: 

Would this be useful on the status command?

nemski:
Probably but I'm not looking to tackle status yet, just validation.

stevehodgkiss:
@nemski with https://github.com/envato/stack_master/pull/74 we're able to execute validate on a list of stacks. What do you think about adding a `--changed` global option to filter stack definitions down to the ones that have changed. That would mean you could do `stack_master diff --changed us-east-1` or even `stack_master validate --changed` to validate all changed stacks. We'd need to improve the output of validate to reflect the stacks that are considered valid though.

nemski:
The problem with this approach is you would need to pass the exit code along and abort if any failed.

I'm about to put up another PR to exit code 1 if validation fails.

stevehodgkiss:
Yep, that would introduce some additional complexity. Do-able though.

features/list.feature
```diff
@@ -0,0 +1,34 @@
...
+      | REGION    | STACK_NAME|
+      | ----------|---------- |
+      | us-east-1 | stack1    |
+      | us-east-1 | stack2    |
+      | us-east-1 | stack3    |
+
+    And the exit status should be 0
+
+  Scenario: Run list command and get machine-readable list of stacks
+    When I run `stack_master list --machine-readable` interactively
```
stevehodgkiss: `interactively` isn't required here...

stevehodgkiss:
What's the status of this PR? Do we still want this behaviour? Should machine readable be the default?

nemski:
Not required right now, but might be in the future. Will close for now.

PR #77, opened 2016-02-08
Fail when validation fails

nemski:
I tried to put some tests together for this, but we'd need to use spawn to avoid the exception cascading into our tests and then we don't have the ability to stub AWS anymore.

### DIFF
--- a/lib/stack_master/validator.rb
+++ b/lib/stack_master/validator.rb
@@ -13,7 +13,7 @@ module StackMaster
       StackMaster.stdout.puts "valid"
     rescue Aws::CloudFormation::Errors::ValidationError => e
       StackMaster.stdout.puts "invalid. #{e.message}"
-      false
+      failed
     end
 
     private
--- a/lib/stack_master/validator.rb
+++ b/lib/stack_master/validator.rb
@@ -12,7 +12,7 @@ module StackMaster
       StackMaster.stdout.puts "Valid"
     rescue Aws::CloudFormation::Errors::ValidationError => e
       $stderr.puts "Validation Failed"
-      $stderr.puts e.message
+      abort e.message
     end
 
     private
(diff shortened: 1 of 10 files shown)


stevehodgkiss:
If we're executing against multiple stacks (running commands multiple times) this won't work, so we shouldn't call abort from within a command. We could add something to `Command` that allows us to mark it as success/failure then we can decide the exit code further upstream like in `CLI`.

nemski:
@stevehodgkiss I went with a different approach, as we don't have access to the instance of Command from the CLI module. Instead we just collect the return values from perform and check if they are all truthy/not nil

stevehodgkiss:
@nemski PR to improve output of `validate` https://github.com/envato/stack_master/pull/79

lib/stack_master/validator.rb
```diff
@@ -13,6 +13,7 @@ def perform
     rescue Aws::CloudFormation::Errors::ValidationError => e
       $stderr.puts "Validation Failed"
       $stderr.puts e.message
+      false
```
stevehodgkiss: A lot of commands don't care about their return value. I'd prefer it if we added this to command and called `failed` here, then collected `command.perform(..).success?` in CLI to determine if all commands succeeded.

```
    module ClassMethods
      def perform(*args)
        new(*args).tap { |command| command.perform }
      end
    end

    def failed
      @failed = true
    end

    def success?
      @failed != true
    end
```

nemski: Yeah but the Command instance is not currently available from the CLI method, so how do we retrieve the return value?

stevehodgkiss: This change makes the return value the instance of the command instead of the return value of perform.

```
    module ClassMethods
      def perform(*args)
        new(*args).tap { |command| command.perform }
      end
    end
```

nemski:
@stevehodgkiss LGTM

PR #78, opened 2016-02-08
Fix config command line parser

nemski:
The Commander documentation is unclear on this, but you actually need to pass the option type to this command for it to work.


stevehodgkiss:
:+1:

gstamp:
:+1:

PR #79, opened 2016-02-08
Update output so to be “stack_name: valid/invalid\n” for each stack

stevehodgkiss:


lib/stack_master/validator.rb
```diff
@@ -7,11 +7,12 @@ def initialize(stack_definition)
     end
 
     def perform
+      StackMaster.stdout.print "#{@stack_definition.stack_name}: "
```
nemski: I wonder if this should happen in CLI.execute_stacks_command

stevehodgkiss: It's only because this command can output it's result in one line that a print makes sense. It probably wouldn't make sense for apply which does a lot of things.

nemski:
LGTM

PR #80, opened 2016-02-10
Filter stack definitions to based on whether they're considered changed

stevehodgkiss:
Enables stack definitions to be filtered further based on whether they are considered changed. For example `stack_master --changed apply us-east-1`.

Needs some more work/refactoring before merging.


gstamp:
Looking good so far.

PR #82, opened 2016-02-17
Parameter values must be strings

stevehodgkiss:
Converts arrays into strings whether supplied through the resolver array or as a list in YAML.

Ref: https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/apis/cloudformation/2010-05-15/api-2.json#L579


nemski:
:+1:

PR #85, opened 2016-03-07
S3 templates

flyinbutrs:
I took the work @lethalpaga started, and got it to a working place. I also added some logic so it doesn't upload the templates every time, and if the existing template MD5 matches, leaves it in place.


flyinbutrs:
Changed to WIP, working on fixing the rspec tests.

PR #87, opened 2016-03-28
S3 templates 

flyinbutrs:
I took the work @lethalpaga started, and got it to a working place. I also added some logic so it doesn't upload the templates every time, and if the existing template MD5 matches, leaves it in place.

rspec tests are passing now, seems to be ready to merge. I can squash if you want.


flyinbutrs:
Just an FYI, there's also only a few real commits in there, but I merged in the intervening changes so it would merge cleanly.

lib/stack_master/stack_definition.rb
```diff
@@ -7,15 +7,29 @@ class StackDefinition
...
+    end
+
     def template_file_path
-      File.join(base_dir, 'templates', template)
+      File.join(template_dir, template)
+    end
+
+    def files_to_upload
+      require 'pry'
+      binding.pry
```
stevehodgkiss: The debug lines shouldn't be in the PR. Is this method used anywhere? Should we be calling it from https://github.com/envato/stack_master/pull/87/files#diff-cf22d74c8f612a3282d758a02c9d21e7R116 ?

lib/stack_master/command.rb
```diff
@@ -21,5 +21,13 @@ def failed
...
       @failed != true
     end
+
+    def failed
+      @failed = true
+    end
+
+    def success?
+      @failed != true
+    end
```
stevehodgkiss: Looks like duplicate method definitions?

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+        end
+      end
+
+      def put_object(options)
+        s3.put_object(options)
+      end
+
+      def url(options)
+        s3_url(@region, options['bucket'], options['prefix'], options['template'])
```
stevehodgkiss: Could we use parameters instead of an options hash? `def url(bucket:, prefix:, template:)`

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def s3_url(region, bucket, prefix, file)
+        prefix += '/' if prefix
+        if region == 'us-east-1'
+          "https://s3.amazonaws.com/#{bucket}/#{prefix}#{file}"
```
stevehodgkiss: Is this special condition for us-east-1 required?

lib/stack_master/commands/apply.rb
```diff
@@ -87,14 +97,43 @@ def create_stack
...
+      end
+
+      def template_value
+        return proposed_stack.maybe_compressed_template_body unless use_s3?
+        s3.url(@s3_config.merge('template' => @stack_definition.template))
+      end
+
+      def files_to_upload
+        return {} unless use_s3?
+        {@stack_definition.template => @stack_definition.template_file_path}
```
stevehodgkiss: This won't work for anything but json templates. We need to ensure the compiled template is sent for sparkleformation/.rb templates. The method to return a compiled string is currently `proposed_stack.maybe_compressed_template_body`. Also, we probably don't want to use the `.rb` extension for the file on S3 given it would be JSON at that stage.

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,74 @@
...
+      def put_object(options)
+        s3.put_object(options)
+      end
+
+      def url(options)
+        s3_url(@region, options['bucket'], options['prefix'], options['template'])
+      end
+
+      def s3_url(region, bucket, prefix, file)
+        prefix += '/' if prefix
```
stevehodgkiss: Maybe introduce a variable to represent the objects key instead of mutating prefix? `object_key = [prefix, file].compact.join('/')`?

lib/stack_master/stack_definition.rb
flyinbutrs: Sorry about the debug flags, I'll remove them.  I think you're right that it's not called. And its incorrect at this point anyway, I had to change how we sourced the files to upload since I wanted to preserve the full path of the files. just having the filename isn't enough because you need the relative path it should be uploaded to. 99% sure it just needs the files_to_upload in apply rather than this one.

lib/stack_master/command.rb
flyinbutrs: oops.. removing.

lib/stack_master/aws_driver/s3.rb
flyinbutrs: Yes, the s3 endpoint syntax is different for us-east-1. It's the first and "default" s3 location, and I don't think they thought far enough ahead when they created it.

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

lib/stack_master/aws_driver/s3.rb
flyinbutrs: sure, I'll rewrite.

lib/stack_master/aws_driver/s3.rb
flyinbutrs: works for me.

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,70 @@
+require 'digest/md5'
+
+module StackMaster
+  module AwsDriver
+    class S3ConfigurationError < StandardError; end
+
+    class S3
+      attr_accessor :region, :bucket, :prefix, :files
+
+      def initialize(region: nil, bucket: nil, prefix: nil, files: nil)
```
flyinbutrs: @stevehodgkiss I decided that since the S3 region / bucket / prefix are always the same per run, I decided to use the intiialize for the s3 driver to set them, and then use accessors in the rest of the function to get that info. I'm having some trouble writing the rspec test though, since I'm... well... a noob at unit testing. I'm trying to figure it out, but I'm wondering if you know how I should modify those tests? 

They have a number of different "options", which used to be passed specifically to upload_files, but I would prefer to pass to the initializer for the s3 driver. How should I use those different tests as part of different s3_driver#initialize instead of as options passed to the method?

````ruby
  subject(:s3_driver) { StackMaster::AwsDriver::S3.new }
     describe '#upload_files' do
````

Comment (stevehodgkiss) [inline lib/stack_master/aws_driver/s3.rb]: The driver classes in stack master are designed to be analogous to the AWS SDK's Client classes. I think only region should be passed in as the initializer. Otherwise, this is more of a single use UploadFiles command class.  One of the main reasons we have these drivers is to enable us to easily run integration tests with cucumber & aruba without hitting AWS at all by using a fake/test driver (see the CloudFormation test driver for an example).

As for how to test, I'm not sure I understand your question properly. Are you asking how to test the initializer? If so I would say that it is unnecessary, the assignments will be tested indirectly through the upload files method, the only thing that might warrant a test is validating the bucket config.

Comment (flyinbutrs) [inline lib/stack_master/aws_driver/s3.rb]: That clarification sort of nullifies my question.  I'll change the implementation, and then the spec tests can largely stay the same.

**Comment by flyinbutrs** [comment]
Believe I've solved the last problems you had, please take a look. rspec is passing, and I am now passing the computed cf body instead of just the file, so it should work for sparkleformations. 

A simple fixture for sparkleformation would be good though, I'd like to add a unit text to verify it computes the CF before uploading, but I don't know SF at all.

File: lib/stack_master/commands/apply.rb [hunk 8]
```diff
@@ -87,17 +97,53 @@ def create_stack
...
+        return proposed_stack.maybe_compressed_template_body unless use_s3?
+        s3.url(bucket: @s3_config['bucket'], prefix: @s3_config['prefix'], region: @s3_config['region'], template: @stack_definition.template)
+      end
+
+      def files_to_upload
+        return {} unless use_s3?
+        {
+          @stack_definition.template => {
+            :path => @stack_definition.template_file_path,
+            :body => proposed_stack.maybe_compressed_template_body
```
stevehodgkiss: A better name for this method `maybe_compressed_template_body` is `compiled_template_body`.

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      private
+
+      def s3
+        @s3 ||= Aws::S3::Client.new(region: @region)
+      end
+    end
+  end
+end
```
stevehodgkiss: We should really cover S3 uploads in our integration tests (cucumber/aruba). That will involve writing a test driver, using it in test mode and making assertions on it in our cucumber scenarios. I'm happy to help with this or provide guidance if you want.

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,73 @@
...
+          file = file[:path]
+          key = template.dup
+          key.prepend("#{prefix}/") if prefix
+          file_md5 = Digest::MD5.file(file).to_s
+          body_md5 = Digest::MD5.hexdigest(body).to_s
+          s3_md5 = current_objects[key] ? current_objects[key].etag.gsub("\"", '') : nil
+
+          # require 'pry'
+          # binding.pry
+          unless file_md5 == s3_md5 or body_md5 == s3_md5
```
stevehodgkiss: While in this situation `or` works fine I would prefer to use `||`.

I'm not sure the file_md5 (file path hash) check is necessary here? Why would the s3 hash equal the hash of the file path locally?

Maybe this could be as simple as: `next if body_md5 == s3_md5` ?.

lib/stack_master/commands/apply.rb
flyinbutrs: fixed, I think I understand that now...  still figuring out the ins and outs of stack_master.

flyinbutrs: Actually, no such method...  template_body?

stevehodgkiss: I was suggesting we rename the method to compiled_template_body

lib/stack_master/aws_driver/s3.rb
flyinbutrs: Short version, it allows the bucket to contain pretty json and/or be synced up to a bucket separately from stack_master.

Long version, this gets a bit toward my own personal use case... at Logicworks, we have a library of CF templates that we use across a wide variety of projects. We have a CF bucket that we manage and update via CI/CD that is read only across a variety of different AWS users/accounts. So in our case, what we'll be doing is syncing that bucket down to the templates/ folder in stack_master.  As a result, the files in the bucket will not actually be writable most of the time, and we're relying on the MD5 of both places to be equal and prevent stack_master from even attempting to rewrite the contents of the bucket. 

And, since the content of body_md5 is the compressed json, and the contents or the bucket are a form of 'pretty' json, I'd prefer stack_master to accept either the pretty of the compressed version of the json to be acceptable.

how about:

````ruby
next if [body_md5, file_md5].include?(s3_md5)
````

lib/stack_master/aws_driver/s3.rb
flyinbutrs: Guidance would be good...  I'm woefully behind when it comes to automated testing and want to get better.

lib/stack_master/aws_driver/s3.rb
stevehodgkiss: Ah, I see now. So this is checking the S3 file against either the raw template file/JSON, or the compiled JSON. Maybe we can rename those variables to make it more obvious what's happening. `raw_template_md5` and `compiled_template_md5`?

lib/stack_master/commands/apply.rb
flyinbutrs: Ah, would you prefer I rebase on master first, or do this as a separate PR after this is merged? I checked, that method is definitely referenced in changes not on my branch.

lib/stack_master/aws_driver/s3.rb
flyinbutrs: Done.

lib/stack_master/commands/apply.rb
stevehodgkiss: I don't mind, whichever you prefer. Happy to leave that refactoring now to focus on getting this merged.

stevehodgkiss:
Re: an integration test. Take a look at `TestDrivers::CloudFormation`. We'll need a `TestDriver::S3`, which basically just keeps any files in memory so that we can query it in the integration test. See apply.feature for how we're testing against the CloudFormation test driver. Let me know if I can help with this :)

flyinbutrs:
Isn't that "spec/aws_driver/s3_spec.rb"? I didn't write that, @lethalpaga did, but it seems to line up with what you're looking for. Maybe move it to test_driver for consistency?

flyinbutrs:
So I'm looking at the CF Driver, and I sort of understand what it's doing, but I'm not quite sure what to do WRT to S3. Are the Virtus.model's modeling the inputs or the outputs of API calls?

flyinbutrs:
Sorry, the past month has been crazy at work, but I'm hoping to circle back on this this weekend.  Is there anything you can point me to on this? It seems like currently the only two "operation" that happen with S3 are getting the list of files in the bucket with their MD5's and putting new objects up into s3. Are those the only operations that need to be testable?

flyinbutrs:
@stevehodgkiss ?

stevehodgkiss:
> It seems like currently the only two "operation" that happen with S3 are getting the list of files in the bucket with their MD5's and putting new objects up into s3. Are those the only operations that need to be testable?

The idea is that we can integration test (s3 uploads in this case) from the outside... given a stack master with s3 configured, and I apply, these files end up on S3 (the test driver for S3). In order to support this the test driver would implement additional methods to query what has been uploaded, and this is what we would test against in integration tests.

I'd be happy to add the integration test stuff myself if it helps... this branch will need updating with master before it's merge-able though.

Thanks!

flyinbutrs:
Sounds good, I will rebase on master, resolve the issues and then push again.

flyinbutrs:
Ok, this is tested and working with all changes from master.

flyinbutrs:
nudge

stevehodgkiss:
Hey @flyinbutrs, sorry for the late reply, I've been quite busy these past few weeks. I like where this is at, but I'd like to make some modifications and add integration tests. Can you add me as a collaborator to your fork? It'll be easier if we can both contribute to the same branch. I'll make time to work on this within the next 1-2 weeks. Thanks for your work on this feature!

flyinbutrs:
No worries, I know how that goes.  I've added you as a collaborator.  I've got a few more PR's in the works too.

PR #89, opened 2016-03-29
Complain when parameters have nil values

stevehodgkiss:
This will cause the update to fail, so this outputs a message indicating where to check for parameter files.

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -14,6 +14,13 @@ module StackMaster
 
       def perform
         diff_stacks
+        if @proposed_stack.empty_parameters?
+          StackMaster.stderr.puts "Empty/blank parameters detected, ensure values exist for those parameters. Parameters will be read from the following locations:"
+          @stack_definition.parameter_files.each do |parameter_file|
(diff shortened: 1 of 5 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -14,6 +14,13 @@ def initialize(config, stack_definition, options = {})
 
       def perform
         diff_stacks
+        if @proposed_stack.missing_parameters?
+          StackMaster.stderr.puts "Empty/blank parameters detected, ensure values exist for those parameters. Parameters will be read from the following locations:"
+          @stack_definition.parameter_files.each do |parameter_file|
```
simpsora: Why loop rather than `join`?

simpsora:
Nice! :+1:

lib/stack_master/commands/apply.rb
stevehodgkiss: I'm not sure exactly what you mean, how would you write it?

simpsora: Ah, the indentation makes that a bit of a pain.  Never mind.

PR #90, opened 2016-03-30
Use change sets to update stacks

stevehodgkiss:
Implements CloudFormation's new change sets feature. Stack updates will now use change sets rather than the update stack API.

screen shot 2016-03-30 at 4 33 45 pm

https://aws.amazon.com/blogs/aws/new-change-sets-for-aws-cloudformation/


lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -6,6 +6,22 @@ def set_region(region)
...
+      end
+
+      def execute_change_set(options)
+        cf.execute_change_set(options)
+      end
+
+      def delete_change_set(options)
+        cf.delete_change_set(options)
+      end
+
```
taoza: Not delegation?

stevehodgkiss: Fixed in 3dc6504

simpsora:
Nice (and _fast_) work!  I haven't tried it in anger, but looks really good.  :+1:

taoza:
:heart_eyes:

taoza:
Nice :+1:

lib/stack_master/change_set.rb
```diff
@@ -0,0 +1,106 @@
...
+      detail_messages = [target_name]
+      if action_name == 'Replace'
+        detail_messages << "#{detail.target.requires_recreation} requires recreation"
+      end
+      detail_messages << "Change source: #{detail.change_source}"
+      if detail.evaluation != 'Static'
+        detail_messages << "#{detail.evaluation} evaluation"
+      end
+      if detail.causing_entity
+        detail_messages << "Causing entity: #{detail.causing_entity}"
```
gstamp: If wasn't familiar with changesets (which I'm not) I might be confused as to what a `Causing entity` was. Is there some easier to understand terminology we might use?  The documentation says it is: "The identity of the entity that triggered this change".  Maybe we could say `Triggered by:`?

stevehodgkiss: Here's an example. Yeah I think triggered by would be easier to understand, in this case at least.

screen shot 2016-03-31 at 2 45 49 pm

stevehodgkiss: Perhaps we could join change source and causing entity, so that it reads `Change source: ParameterReference.VpcAz1` ?

gstamp:
Code looks good. I don't have any outstanding CF changes to test it with unfortunately.

:+1:

stevehodgkiss:
@gstamp I'm just playing with the example checked into the repo to test it out - https://github.com/envato/stack_master/tree/master/example/simple

Issue #81, opened 2016-02-10
Add support for more template generators


stevehodgkiss:
- YAML
- `cfndsl` - https://github.com/stevenjack/cfndsl

Any others?

gstamp:
It might be nice to have JSON support with comments. Would need to find a library to strip them out I guess.

Another one might be just using plain old ERB?

flyinbutrs:
If you want JSON support with comments, technically you can just use YAML.  An alternate syntax of YAML is actually just JSON + comments, so any YAML parser should be able to parse a JSON file with comments in it.

````ruby
[1] pry(main)> require 'yaml'
=> true
[2] pry(main)> test = "{\n#testing\n}"
=> "{\n#testing\n}"
[3] pry(main)> YAML.load(test)
=> {}
````

So yeah, +1 for YAML input. should be pretty trivial to add, I'll fork and see.

neillturner:
cfndsl support would be really good as it is the most popular generator of cloudformation templates that is written in ruby. it should be really easier to integrate as it is just running cfndsl commands. 
I also use the ppjson to pretty print the generated cloudformation code. 

I've built alot of examples for cfndsl at https://github.com/neillturner/cfndsl_examples
and there is a utility to convert templates to cfndsl format that is very handy.

thekindofme:
Hi @neillturner we have added basic CfnDsl template support with #99. However, this dose not add support for using cfndsl variables. How would you like that supported in SM? How would you imagine a user supply values for cfndsl variables when using a cfndsl template with SM?

PR #91, opened 2016-03-31
Make it easier to add support for new template types

thekindofme:
This work will allow us to easily add support for new template types, such as YML (as requested in https://github.com/envato/stack_master/issues/81)


lib/stack_master/template_compiler.rb
```diff
@@ -1,21 +1,25 @@
...
-        template_body = File.read(template_file_path)
-        if template_body.size > MAX_TEMPLATE_SIZE
-          # Parse the json and rewrite compressed
-          JSON.dump(JSON.parse(template_body))
-        else
-          template_body
-        end
-      end
+    def self.file_ext(template_file_path)
+      File.extname(template_file_path).gsub('.', '').to_sym
```
stevehodgkiss: What about a file name with multiple extensions? Would it be useful to support that? `my_template.erb.json` for example.

stevehodgkiss:
Much better :+1:

lib/stack_master/template_compiler.rb
thekindofme: We could add support for that if we need it, in the future? (YAGNI) or we could do it now. What do you think?

stevehodgkiss: Let's do it later if we need it.

PR #95, opened 2016-04-01
Add support for YML templates

thekindofme:
As discussed in https://github.com/envato/stack_master/issues/81


spec/fixtures/templates/yml/invalid_myapp_vpc.yml
```diff
@@ -0,0 +1,6 @@
+---
+  Descr : : iption: "A test VPC template"
+this is not valid yml
+  Resources:
+    Vpc:
+      Type: "AWS::EC2::VPC"
```
stevehodgkiss: Is this file used anywhere?

thekindofme: It's not :) removing it.

stevehodgkiss: It could make a good test case :)

thekindofme: That's how it was used before. However, I (should have) removed it together
with the error handling code in Yaml.rb as I moved the error handling to
TemplateCompiler.

lib/stack_master/template_compilers/yaml.rb
```diff
@@ -0,0 +1,12 @@
+module StackMaster::TemplateCompilers
+  class Yaml
+
+    def self.compile(template_file_path)
+      template_body = File.read(template_file_path)
+      JSON.dump(YAML.load(template_body))
+    end
+
+    StackMaster::TemplateCompiler.register(:yaml, self)
+    StackMaster::TemplateCompiler.register(:yml, self)
```
nemski: Does the `default_template_compilers` option not make this redundant?

nemski:
:+1:

lib/stack_master/template_compilers/yaml.rb
stevehodgkiss: Yeah I think only the `:yaml` definition would be required... The file extension mapping is handled in config.

thekindofme: Yep it's not needed

stevehodgkiss:
:+1:

PR #96, opened 2016-04-01
Handle Template Compilation Errors

thekindofme:
Right now the application would just crash if there is an template compilation error. This PR adds some error handling that deals with template compilation errors, so the application will output an error and shutdown gracefully if a template compilation error was to happen.

### DIFF
--- a/lib/stack_master/command.rb
+++ b/lib/stack_master/command.rb
@@ -35,7 +35,7 @@ module StackMaster
 
     def error_message(e)
       msg = "#{e.class} #{e.message}"
-      msg = "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
+      msg << "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
       msg
     end
 
--- a/lib/stack_master/command.rb
+++ b/lib/stack_master/command.rb
@@ -22,7 +22,7 @@ module StackMaster
         catch(:halt) do
           super
         end
-      rescue Aws::CloudFormation::Errors::ServiceError => e
+      rescue Aws::CloudFormation::Errors::ServiceError, TemplateCompiler::TemplateCompilationFailed => e
         failed "#{e.class} #{e.message}"
       end
     end
--- a/lib/stack_master/command.rb
+++ b/lib/stack_master/command.rb
@@ -23,7 +23,7 @@ module StackMaster
           super
         end
       rescue Aws::CloudFormation::Errors::ServiceError, TemplateCompiler::TemplateCompilationFailed => e
-        failed "#{e.class} #{e.message}"
+        failed error_message(e)
       end
     end
 
@@ -32,6 +32,12 @@ module StackMaster
     end
 
     private
+
+    def error_message(e)
+      msg = "#{e.class} #{e.message}"
+      msg = "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
+      msg
+    end
 
     def failed(message = nil)
       StackMaster.stderr.puts(message) if message
--- a/lib/stack_master/template_compiler.rb
+++ b/lib/stack_master/template_compiler.rb
@@ -4,7 +4,7 @@ module StackMaster
 
     def self.compile(config, template_file_path)
       template_compiler_for_file(template_file_path, config).compile(template_file_path)
-    rescue Exception
+    rescue StandardError
       raise TemplateCompilationFailed.new("Failed to compile #{template_file_path}.")
     end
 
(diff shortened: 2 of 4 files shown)


lib/stack_master/template_compiler.rb
```diff
@@ -1,8 +1,11 @@
 module StackMaster
   class TemplateCompiler
+    class TemplateCompilationFailed < RuntimeError; end
 
     def self.compile(config, template_file_path)
       template_compiler_for_file(template_file_path, config).compile(template_file_path)
+    rescue Exception
+      raise TemplateCompilationFailed.new("Failed to compile #{template_file_path}.")
```
stevehodgkiss: Why rescue Exception over just StandardError?

lib/stack_master/template_compiler.rb
```diff
@@ -1,8 +1,11 @@
 module StackMaster
   class TemplateCompiler
+    class TemplateCompilationFailed < RuntimeError; end
```
stevehodgkiss: Style nitpick. We usually write this like so: `TemplateCompilationFailed = Class.new(RuntimeError)`

spec/stack_master/command_spec.rb
```diff
@@ -45,4 +45,13 @@ def perform
       expect { command_class.perform(error_proc) }.to output(/the message/).to_stderr
     end
   end
+
+  context 'when a CF error occurs' do
```
stevehodgkiss: This would be a template compile error, not a cloudformation error?

lib/stack_master/template_compiler.rb
thekindofme: Agreed. StandardError should be more than enough. Exception is too broad.

spec/stack_master/command_spec.rb
thekindofme: Copypasta error :)

lib/stack_master/template_compiler.rb
```diff
@@ -4,7 +4,7 @@ class TemplateCompiler
 
     def self.compile(config, template_file_path)
       template_compiler_for_file(template_file_path, config).compile(template_file_path)
-    rescue Exception
+    rescue StandardError
```
stevehodgkiss: `StandardError` can be omitted because it's the default when using `rescue` without an error class

stevehodgkiss:
Nice work! :+1:

lib/stack_master/template_compiler.rb
```diff
@@ -1,8 +1,11 @@
 module StackMaster
   class TemplateCompiler
+    TemplateCompilationFailed = Class.new(RuntimeError)
 
     def self.compile(config, template_file_path)
       template_compiler_for_file(template_file_path, config).compile(template_file_path)
+    rescue StandardError
```
nemski: Can we be more explicit here?

I think it would be preferable if each template compiler handled the relevant exceptions and re-raised them as StackMaster exceptions.

stevehodgkiss: The intention is to be more generic here, each new template compiler will have it's own errors, and if they failed (in one way or another), we wrap in our SM compile error and can handle it more generically upstream (in cli.rb)...

@thekindofme we should make sure the errors `.cause` message is being outputted to STDERR. Otherwise it won't be obvious why compilation failed. Cause is the wrapped error, automatically provided by ruby.

stevehodgkiss: What happens when an error is wrapped in Ruby is, the original error is available in the wrapped errors `#cause`... so we could check and use that (if it's available) in the error message we display to the user.

thekindofme: @stevehodgkiss @nemski We don't print out the 'cause' right now. I'm wondering if we should include more than just the cause error's message in our output. May be `cause.inspect` would be better?

message:
```
"(<unknown>): mapping values are not allowed in this context at line 1 column 8"
```

Inspect:
```
#<Psych::SyntaxError: (<unknown>): mapping values are not allowed in this context at line 1 column 8>
```

stevehodgkiss: Yeah we could use either `#inspect`, but I think that looks too low level (use of angled brackets). It would be better to build our own error message based on the properties of the error. `e.class.name` + `e.message` could be enough (or using `e.cause` if it's not nil)?

thekindofme: @stevehodgkiss  I've added some code to display (as suggested in the last reply) the cause of an exception (if present), in Command.perform.

lib/stack_master/command.rb
```diff
@@ -33,6 +33,12 @@ def success?
 
     private
 
+    def error_message(e)
+      msg = "#{e.class} #{e.message}"
+      msg = "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
```
stevehodgkiss: Should this be `msg << "..."` ?

thekindofme: :) yes it should be

stevehodgkiss:
:+1:

PR #98, opened 2016-04-05
Make template compilers configurable by user

thekindofme:
## Context
We are going to add support for [Cfndsl](https://github.com/stevenjack/cfndsl) in a future PR. And this means that there is going to be two possible template compilers for the .rb files. Hence we need template compiler to file type mapping, to be configurable.

## Change
Let users configure what template compiler is used for a given file extension.

Related documentation will be added in the PR that introduce Cfndsl.

### DIFF
--- a/lib/stack_master/config.rb
+++ b/lib/stack_master/config.rb
@@ -42,7 +42,7 @@ module StackMaster
       end
       @region_defaults = normalise_region_defaults(config.fetch('region_defaults', {}))
       @stacks = []
-      @template_compilers = default_template_compilers
+      load_template_compilers(config)
       load_config
     end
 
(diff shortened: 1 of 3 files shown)


lib/stack_master/config.rb
```diff
@@ -62,6 +62,21 @@ def unalias_region(region)
...
+      populate_template_compilers(config.fetch('template_compilers', {}))
+      merge_defaults_to_user_defined_compilers
+    end
+
+    def merge_defaults_to_user_defined_compilers
+      @template_compilers = default_template_compilers.merge(@template_compilers)
+    end
+
+    def populate_template_compilers user_defined_compilers
+      user_defined_compilers.each do |key, val|
```
thekindofme: I've decided to have everything in the map as symbols. So it's more structured and also matches how we do things already.

spec/fixtures/stack_master.yml
```diff
@@ -4,6 +4,8 @@ region_aliases:
 stack_defaults:
   tags:
     application: my-awesome-blog
+template_compilers:
+  rb: ruby_dsl
```
stevehodgkiss: Why call it `ruby_dsl` over `cfndsl` ?

lib/stack_master/config.rb
```diff
@@ -62,6 +62,21 @@ def unalias_region(region)
     end
 
     private
+    def load_template_compilers config
```
stevehodgkiss: No curly brackets around method parameters?

spec/fixtures/stack_master.yml
thekindofme: This was put in solely as test data (and intentionally avoided any connection with cfndsl, as this could be anything and the code we test wouldn't care about it). Should we use cfndsl instead?

stevehodgkiss: Ah I see... cool! No `ruby_dsl` is fine :)

stevehodgkiss:
:+1:

PR #99, opened 2016-04-05
Add CfnDsl Template Compliation Support

thekindofme:
## Context
We want to support CfnDsl. Refer: https://github.com/envato/stack_master/issues/81

CfnDsl: https://github.com/stevenjack/cfndsl

## Change
Use the cfndsl gem internally to compile any ".rb" templates.

## Considerations
We don't support passing in variables to CfnDsl templates, yet.


stack_master.gemspec
```diff
@@ -36,4 +36,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency "table_print"
   spec.add_dependency "dotgpg"
   spec.add_dependency "deep_merge"
+  spec.add_dependency "cfndsl"
```
thekindofme: Currently we have hard dependencies on sparkleformation and cfndsl, although some users of SM will never use them or the associated functionality. I wonder if we could not have a hard dependency on them, check for them if the user attempt to use them, then complain and fail if they don't have them (+ask them to install them first if they want to use whatever template they attempted to). Thoughts?

stevehodgkiss: That sounds reasonable to me.

stevehodgkiss:
:+1:

stack_master.gemspec
thekindofme: I'll do the above mentioned change in a seperate PR.

stevehodgkiss: Cool. We'll also need to update the readme about the fact that they need to be installed separately...

thekindofme: Yep. That too. Noted.

thekindofme: It seems that we will need to "break out" of bundler to be able to load gem X without having that in our gemspec. I think at this point it's not worth the cost. I suggest we keep doing what we are doing already (having hard dependencies).

If we want to pursue this path though, I could look at using something like [this](https://github.com/janlelis/debundle.rb/blob/modern/debundle.rb) That'll allow us to break out of bundler jail. Which might invite other undesirable issues. Thoughts?

PR #101, opened 2016-04-06
Remove Hard Dependencies on Template Compilation Related Gems

thekindofme:
## Context

> Currently we have hard dependencies on sparkleformation and cfndsl, although some users of SM will never use them or the associated functionality. I wonder if we could not have a hard dependency on them, check for them if the user attempt to use them, then complain and fail if they don't have them (+ask them to install them first if they want to use whatever template they attempted to). 

Refer the conversation at: https://github.com/envato/stack_master/pull/99

### DIFF
--- a/lib/stack_master/template_compilers/cfndsl.rb
+++ b/lib/stack_master/template_compilers/cfndsl.rb
@@ -2,9 +2,19 @@ module StackMaster::TemplateCompilers
   class Cfndsl
 
     def self.compile(template_file_path)
-      require 'cfndsl'
+      load_cfndsl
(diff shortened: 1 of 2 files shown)


thekindofme:
@stevehodgkiss  It seems that we will need to "break out" of bundler to be able to load gem X without having that in our gemspec. I think at this point it's not worth the cost. I suggest we keep doing what we are doing already (having hard dependencies).

If we want to pursue this path though, I could look at using something like this That'll allow us to break out of bundler jail. Which might invite other undesirable issues. Thoughts?

thekindofme:
It was decided that this is not worth the effort or the potential future consequences.

PR #103, opened 2016-04-13
Ensure that aws-sdk must be newer than 2.2.31

joneslee85:
according to CHANGELOG of aws-sdk, the CreateChangeSet function was introduced since 2.2.31

### DIFF
--- a/stack_master.gemspec
+++ b/stack_master.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ruby-progressbar"
   spec.add_dependency "commander"
   spec.add_dependency "virtus"
-  spec.add_dependency "aws-sdk", ">= 2.2.31"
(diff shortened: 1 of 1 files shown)


gstamp:
/cc @stevehodgkiss 

👍 from me though.

stack_master.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ruby-progressbar"
   spec.add_dependency "commander"
   spec.add_dependency "virtus"
-  spec.add_dependency "aws-sdk", "~> 2.1"
+  spec.add_dependency "aws-sdk", ">= 2.2.31"
```
stevehodgkiss: We should probably still pin to the minor version. Would `~> 2.2.31` work?

stevehodgkiss:
ping @joneslee85

Issue #102, opened 2016-04-11
Way confusing error message


andrewjhumphrey:
If you define a parameter this way:

```
  parameters.enable_dns_hostnames do
    description 'Enable DNS hostnames on the VPC?'
    type 'String'
  end
```

And then in the parameters file use a boolean type by mistake:

```
enable_dns_hostnames: true
```

instead of 

```
enable_dns_hostnames: `true`
```

You get the not so awesome error message:

```
error: expected params[:parameters][0][:parameter_value] to be a string. Use --trace to view backtrace
```

And are left to your own devices to figure out which parameter has the broken type.

Would be nice if it at least gave a hint,  even something like 'the parameter starting with the letter b has caused us problems' would be better than the current message ;-P

stevehodgkiss:
PR https://github.com/envato/stack_master/pull/107

patrobinson:
Closing due to age.

PR #107, opened 2016-04-20
Convert boolean parameter values to strings

stevehodgkiss:
Addresses part of #102


nemski:
So all parameters should be strings?

redterror:
Any reason not to merge this?  Seems simple / straightforward.

redterror:
Thanks!

PR #108, opened 2016-04-22
Create a LatestAmi Parameter Resolver

nemski:
When trying to use an AMI that is shared (publicly or privately) the tags of the AMI are not visible.
So we must resort to other attributes, such as the name, to find the required AMI.

This implements a new resolver and de-couples the logic from LatestAmiByTag into a separate class.

We also do some validation that the AMI is from the correct owner. ~~Users need the `iam:ListUsers` permission to do this. This might be a breaking change for some users! However~~ currently the only reason LatestAmiByTags won't return AMIs from other owners is because, the tags of the AMI are not visible. LatestAmi should, in my opinion, only return results from the current account by default. Doing otherwise would be surprising to users and a potential security risk.


lib/stack_master/parameter_resolvers/ami_finder.rb
```diff
@@ -0,0 +1,43 @@
...
+        sorted_images = images.sort do |a, b|
+          Time.parse(a.creation_date) <=> Time.parse(b.creation_date)
+        end
+        sorted_images.last
+      end
+
+      private
+
+      def aws_account_id
+        arn = iam.list_users.users.first.arn
```
simpsora: Unfortunately, this wouldn't work in a scenario when the account didn't have any users.  There are certainly AWS accounts which delegate user management to other systems (via federation or role assumption).

Alas, there isn't a canonical way to find the current account's account number.

However, there _is_ a rather roundabout way to get it:
```
$ aws elastictranscoder list-presets --query 'Presets[0].Arn' | cut -d: -f5
XXXXXXXXXXXX
```

Obviously the caller needs permissions to call `elastictranscoder:listPresets` for that to work.

lib/stack_master/parameter_resolvers/latest_ami.rb
```diff
@@ -0,0 +1,18 @@
+module StackMaster
+  module ParameterResolvers
+    class LatestAmi < Resolver
+      array_resolver class_name: 'LatestAmi'
```
stevehodgkiss: This should be a different name to what the singular class name is. Array resolvers work like this:

```
latest_amis:
  - tag:application=blah
  - tag:name=blah
```

and the resulting parameter should be a comma separated list of AMI's `ami-123,ami-456`.

simpsora:
Love it!

Also agree 100% with your assertion that `LatestAmi` should only return images belonging to the current account.

simpsora:
May also be worth documenting other valid values for the `owner` parameter:

```
Valid values: amazon | aws-marketplace | self | AWS account ID | all
```

(from http://docs.aws.amazon.com/AWSEC2/latest/CommandLineReference/ApiReference-cmd-DescribeImages.html)

lib/stack_master/parameter_resolvers/ami_finder.rb
nemski: I wonder if we can just use `self` as you described below.

lib/stack_master/parameter_resolvers/latest_ami.rb
nemski: Fixed in ad9062b

nemski:
@simpsora @stevehodgkiss I've made it so this now doesn't require the user to have specific permissions, we can just refer to "self" as the account ID by default.

simpsora:
Looks good @nemski! :+1:

nemski:
@simpsora we can't use `amazon | aws-marketplace | self | all` as the owner (from parameters file) because we're passing it into the filter as well as the method call. That's a limitation in how I've implemented it, but I can't see another way to allow users to specify a separate parameter without potentially causing a namespace clash.

lib/stack_master/parameter_resolvers/ami_finder.rb
```diff
@@ -0,0 +1,34 @@
...
+    class AmiFinder
+      def initialize(region)
+        @region = region
+        @owner = ['self']
+      end
+
+      def build_filters(value, prefix = nil)
+        value.split(',').map do |name_with_value|
+          name, value = name_with_value.strip.split('=')
+          @owner.push name if name == 'owner_id'
```
simpsora: Should this really `push`?  If I specified an owner value, I'd expect the results to be restricted to that owner.  Leaving `self` in the list of possible owners might return more than what the caller was expecting.

lib/stack_master/parameter_resolvers/ami_finder.rb
```diff
@@ -0,0 +1,34 @@
...
+          name = prefix ? "#{prefix}:#{name}" : name
+          puts value.class
```
stevehodgkiss: Debug line needs removing...

spec/stack_master/parameter_resolvers/latest_ami_by_tag_spec.rb
```diff
@@ -1,33 +0,0 @@
...
-  context 'when no matches are found' do
-    before do
-      ec2.stub_responses(:describe_images, images: [])
-    end
-
-    it 'returns nil' do
-      expect(resolver.resolve('my-tag=my-value')).to be_nil
-    end
-  end
-end
```
stevehodgkiss: Does this spec no longer work as expected?

stevehodgkiss:
Looks good! :)

spec/stack_master/parameter_resolvers/latest_ami_by_tag_spec.rb
nemski: Nope, all the logic now lives in the `ami_helper` so specs are moved to that class.

stevehodgkiss: How do we know that this resolver still works?

simpsora:
@nemski even with your comment above, I'm :+1: to this change.  We can address it in the future if it becomes a problem.

lib/stack_master/parameter_resolvers/ami_finder.rb
nemski: I was expecting `describe_images` to AND the owner and filter, but owner over-rides the filter.

spec/stack_master/parameter_resolvers/latest_ami_by_tag_spec.rb
nemski: Created some integration tests in 766f5a7

lib/stack_master/parameter_resolvers/latest_ami.rb
nemski: Actually fixed in db8e9a3

lib/stack_master/parameter_resolvers/ami_finder.rb
simpsora: What I mean is that `@owner` is initialized with `['self']`, and then here you're pushing onto `@owner`, so it now contains two values.  If it overrides the filter, then the end result of this will use both `self` and the passed-in `owner_id` (if I'm not missing something).

README.md
```diff
@@ -211,6 +211,17 @@ web_ami:
...
+### Latest AMI by attribute
+
+Looks up the latest AMI ID by a given set of attributes. This selects the latest wily hvm AMI from Ubuntu (using the account id):
+
+```yaml
+bastion_ami:
+  latest_amis: owner_id=099720109477,name=ubuntu/images/hvm/ubuntu-wily-15.10-amd64-server-*
+```
+
+By default only AMIs from the current account are returned.
```
stevehodgkiss: Maybe add a note about how to override this?

README.md
```diff
@@ -211,6 +211,17 @@ web_ami:
 
 Note that the corresponding array resolver is named `latest_amis_by_tags`
 
+### Latest AMI by attribute
+
+Looks up the latest AMI ID by a given set of attributes. This selects the latest wily hvm AMI from Ubuntu (using the account id):
+
+```yaml
+bastion_ami:
+  latest_amis: owner_id=099720109477,name=ubuntu/images/hvm/ubuntu-wily-15.10-amd64-server-*
```
stevehodgkiss: It might be worth linking to the documentation to find out other things to filter by - `https://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#describe_images-instance_method`.

lib/stack_master/parameter_resolvers/ami_finder.rb
```diff
@@ -0,0 +1,36 @@
...
+        owner_id = []
+        filters = value.split(',').map do |name_with_value|
+          name, value = name_with_value.strip.split('=')
+          owner_id.push value if name == 'owner_id'
+          name = prefix ? "#{prefix}:#{name}" : name
+          { name: name, values: [value] }
+        end
+        @owner = owner_id unless owner_id.empty?
+        filters
+      end
```
stevehodgkiss: Related to the earlier discussion about owner... I think the `owners` option (not the filter) sets the scope of what the filters can ... filter. Because we're setting a default of `['self']` I can see why we need to broaden the scope of owner to make the filter work.

How would you select an image owned by amazon or in the aws marketplace with this resolver? I can't see how it would be possible... Is there a way to lift the `self` restriction?

lib/stack_master/parameter_resolvers/ami_finder.rb
nemski: Fixed in b258da7

lib/stack_master/parameter_resolvers/ami_finder.rb
nemski: > I think the owners option (not the filter) sets the scope of what the filters can ... filter

Correct

> How would you select an image owned by amazon or in the aws marketplace with this resolver?

The only way at the moment would be to use the account ID.

I can't think of a better way to do it and since it's not possible at the moment I still think we've made an improvement.

README.md
nemski: Added in e225df2

lib/stack_master/parameter_resolvers/ami_finder.rb
stevehodgkiss: It'd be better to name the variables `@owners` and `owner_ids` since we're dealing with arrays and the first one then matches up with the api call a bit better.

spec/stack_master/parameter_resolvers/ami_finder_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        expect(resolver.build_filters('my-tag=my-value', 'tag')).to eq [
+          { name: 'tag:my-tag', values: ['my-value']}
+        ]
+      end
+    end
+
+    context 'owner_id is specified' do
+      it 'updates the owner instance variable' do
+        resolver.build_filters('owner_id=0123456789', nil)
+        expect(resolver.instance_variable_get('@owner')).to eq ['0123456789']
```
stevehodgkiss: It would be better to assert that the API was called with the expected value, rather than check the internal state of the object itself.

stevehodgkiss: For example `expect(ec2).to receive(:describe_images).with(owners: [...], filters: anything)`

nemski:
@stevehodgkiss as discussed yesterday I've updated latest_ami to take a hash.

This simplifies how we pass the owner_id to the describe_images method and allows us to use the keywords mentioned by @simpsora

lib/stack_master/parameter_resolvers/latest_ami.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def resolve(value)
+        case value
+        when String
+          owners = ['self']
+          filters = @ami_finder.build_filters_from_string(value)
+        when Hash
+          owners = Array(value.delete('owner_id').to_s || 'self')
+          filters = @ami_finder.build_filters_from_hash(value)
+        end
```
stevehodgkiss: Why support both string and hash options?

I'm not sure what value the hash options gives in this implementation, because the hash and the key value string are essentially the same thing, the string just needs parsing into a hash. I thought we wanted to use a hash to be able to nest a level deeper than the key value string would allow.

nemski: We can strip out string easily.

Were you envisaging it look like this?
```yaml
owners: 123456
filters:
  name: foo*
```

stevehodgkiss: Yeah I thought that was the reason for wanting to move to a hash... The way it is now could just as easily be done with a key value string like our other resolver inputs. We'd just need to parse the key value string into a hash first, the rest is the same logic.

nemski: Done in dccb245

nemski:
Ping @stevehodgkiss

lib/stack_master/parameter_resolvers/latest_ami.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+        @ami_finder = AmiFinder.new(@stack_definition.region)
+      end
+
+      def resolve(value)
+        owners = Array(value.fetch('owners', 'self').to_s)
+        filters = @ami_finder.build_filters_from_hash(value['filters'])
```
stevehodgkiss: It'd be better to use `value.fetch('filters')` so that a more obvious error is raised when filters is missing.

stevehodgkiss:
👍

taoza:
@stevehodgkiss When are we going to release this feature?

nemski:
@taoza I released this today as version 0.5.0

taoza:
> @taoza I released this today as version 0.5.0

@nemski Works like a charm! Thanks :)

PR #111, opened 2016-05-11
Fix SNS topic name doc

taoza:
The correct parameter resolver should be sns_topic_name rather than
sns_topic

/cc @stevehodgkiss


stevehodgkiss:
👍

PR #115, opened 2016-06-21
Add region_alias support for multiple aliases in a single region

flyinbutrs:
Multiple region_aliases can now be in the same region, and with
the prepend_region_alias_to_stack_names flag set to true, the region_alias
will be part of the stack name for easier readability.

```
prepend_region_alias_to_stack_names: true
region_alias:
  - production: us-east-1
stacks:
  production:
    vpc:
      template: vpc.json
```

The config above would create a stack named "production-vpc"

It would also resolve parameters from:

parameters/vpc.json
parameters/us-east-1/vpc.json
parameters/production/vpc.json

### DIFF
--- a/lib/stack_master/stack_definition.rb
+++ b/lib/stack_master/stack_definition.rb
@@ -4,7 +4,9 @@ module StackMaster
 
     values do
       attribute :region, String
+      attribute :region_alias, String
       attribute :stack_name, String
+      attribute :raw_stack_name, String
       attribute :template, String
       attribute :tags, Hash
       attribute :notification_arns, Array[String]
@@ -19,7 +21,12 @@ module StackMaster
     end
 
     def parameter_files
-      [ default_parameter_file_path, region_parameter_file_path ] + additional_parameter_lookup_file_paths
+      [
+        default_parameter_file_path,
+        alias_parameter_file_path,
+        region_parameter_file_path,
+        additional_parameter_lookup_file_paths
+      ].flatten.uniq
     end
 
     def stack_policy_file_path
@@ -34,6 +41,10 @@ module StackMaster
       end
     end
 
+    def alias_parameter_file_path
+      File.join(base_dir, 'parameters', "#{region_alias}", "#{underscored_stack_name}.yml")
+    end
+
     def region_parameter_file_path
       File.join(base_dir, 'parameters', "#{region}", "#{underscored_stack_name}.yml")
     end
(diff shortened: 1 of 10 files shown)


lib/stack_master/stack_definition.rb
```diff
@@ -19,7 +21,12 @@ def template_file_path
     end
 
     def parameter_files
-      [ default_parameter_file_path, region_parameter_file_path ] + additional_parameter_lookup_file_paths
```
flyinbutrs: FYI, I'm not entirely clear on the purpose of "additional_parameter_lookup_file_paths" vs as I did it below, explicitly resolving explicitly to the region_alias as I did below. I've preserved both, but I'm interested in the rationale behind additional_parameter_lookup_file_paths.

stacktemplates/stack_master.yml.erb
```diff
@@ -1,3 +1,7 @@
+# prepend_region_alias_to_stack_names: true
+# region_aliases:
+# stack_defaults:
+#   tags:
```
flyinbutrs: I felt the default stack_master.yml from the init function was a little too barren, so I added some extra empty (commented out) options. Is that ok?

stevehodgkiss:
Will this break how we currently define region aliases (as a hash)? It would be great if we could handle both to avoid having to upgrade previously written stack_master.yml files..

flyinbutrs:
I think that was a typo in my example, but let me check. I can also rebase this now that s3 is merged in. And in the next week or so I should be able to work out the test failures.

l-arjun-rao:
Hi @flyinbutrs and @stevehodgkiss  
I chanced upon this pull-request, and this feature would really be useful to me too, as I need the ability to prefix the stack-names with prod/dev/test etc.
Just thought I'd touch base to ask if you were able to work out the test failures etc. 

A big +1 from my side for this feature :-)

et304383:
I'd honestly rather see a breaking change to just get rid of stack master using regions as keys or aliases.

I have never seen any project deal in terms of "regions" for separate stacks and I've always seen stacks separated in terms of environment (dev, staging, prod).

I highly suggest that stack master be overhauled to kill the region alias key in a major breaking release and move to using environments.  Make region just an attribute of the stack so that an environment can be multi region if needed.

patrobinson:
> I highly suggest that stack master be overhauled to kill the region alias key in a major breaking release and move to using environments. Make region just an attribute of the stack so that an environment can be multi region if needed.

I agree, but I think we need to come up with a better way of handling "environments" before that can happy. Open to opinions!

patrobinson:
This seems to have been super-seeded by #221

patrobinson:
This PR is quite stale and has been superseeded. Closing for now but re-open when/if you're ready.

Issue #104, opened 2016-04-14
Ability to pass user data as a parameter.


mhussain:
Requesting the ability to pass user data as parameter to the stack instead of having to write in inline in the template.

Something like

``` yaml
DateInit: files/user_data.txt
```

``` json
"DataInit": {
  "Type": "String",
  "Description": "Initialisation script"
}
...

"UserData": { "Fn::Base64": { "Ref": "DataInit" } }

...
```

simpsora:
:+1: to this idea -- it would be nice to have UserData content extracted to a separate file.

For this to be most useful, the content of the would still need to be parsed by CloudFormation, in order to resolve intrinsic functions (`Fn::Ref`, `Fn::GetAtt`, etc.).  Use of those functions may introduce some strange requirements when it comes to quoting the content.

stevehodgkiss:
I saw an interesting solution to this in a sparkleformation issue: https://github.com/sparkleformation/sparkle_formation/issues/54#issuecomment-143136818

It enables separate user data files while being able to use ERB to call sparkle functions inside that file like `<%= region! %>`.

simpsora:
@stevehodgkiss that's an interesting approach, although it's a shame that it only works with Sparkle.  I'd love to see a more generic approach if possible (though I don't currently see how).

`cloud-init` has quite a bit more functionality than just user-data shell scripts, see here:
https://help.ubuntu.com/community/CloudInit

About the only idea I have is to create a grammar of some kind to allow embedding CF references/functions inline in the script, then replacing them with closequote-var-openquote:
```...", { "Ref" : "Foo" }, "...```

**Comment by simpsora** [comment]
@mhussain: 0.6.0 has been released, which contains `user_data_file!`.  See https://github.com/envato/stack_master/pull/117 for details.

**Comment by stevehodgkiss** [comment]
I think the approach/code used in #117 could be used in JSON/YAML templates quite easily too, we'd just need to run it through the same Erubis context. It could look like this:

```
  "UserData": <%= user_data_file!('app.erb') %>
```

PR #117, opened 2016-07-14
Add support for user data files in sparkleformation templates

stevehodgkiss:
Based on the code in this comment https://github.com/sparkleformation/sparkle_formation/issues/54#issuecomment-143136818

With this change, you can write the following in templates:

```
  user_data user_data_file!('web.erb', role: :web)
```

It will look in templates/user_data/* for user data files, read the template and compile it with each line being a separate string in the final array. The files can look like this and make use of SF helpers:

```
 #!/bin/bash
 
REGION=<%= region! %>
ROLE=<%= has_var?(:role) ? role : 'hmm' %>
echo $REGION
```

Renders like this:

```
+        "UserData": {
+          "Fn::Base64": {
+            "Fn::Join": [
+              "",
+              [
+                "#!/bin/bash\n",
+                "\n",
+                "REGION=",
+                {
+                  "Ref": "AWS::Region"
+                },
+                "ROLE=web",
+                "\n",
+                "echo $REGION\n"
+              ]
+            ]
+          }
+        }
+
```

Ref: https://github.com/envato/stack_master/issues/104

### DIFF
--- a/lib/stack_master/sparkle_formation/user_data_file.rb
+++ b/lib/stack_master/sparkle_formation/user_data_file.rb
@@ -22,7 +22,22 @@ class SparkleFormation
       def _user_data_file(file_name)
         file_path = File.join(::SparkleFormation.sparkle_path, 'user_data', file_name)
         template = File.read(file_path)
-        base64!(join!(SfEruby.new(template).evaluate(self)))
+
+        compiled_template = SfEruby.new(template).evaluate(self)
+        formatted_cf_template = compiled_template.flat_map do |lines|
+          if String === lines
+            newlines = []
+            lines.count("\n").times do
+              newlines << "\n"
+            end
+            lines.split("\n").map do |line|
+              "#{line}#{newlines.pop}"
+            end
+          else
+            lines
+          end
+        end
+        base64!(join!(formatted_cf_template))
       rescue Errno::ENOENT => e
         Kernel.raise UserDataFileNotFound, "Could not find user data file at path: #{file_path}"
       end
--- a/lib/stack_master/sparkle_formation/user_data_file.rb
+++ b/lib/stack_master/sparkle_formation/user_data_file.rb
@@ -32,6 +32,39 @@ class SparkleFormation
         end
       end
 
+      # Splits up long strings with multiple lines in them to multiple strings
+      # in the CF array. Makes the compiled template and diffs more readable.
+      class CloudFormationLineFormatter
+        def self.format(template)
+          new(template).format
+        end
+
+        def initialize(template)
+          @template = template
+        end
+
+        def format
+          @template.flat_map do |lines|
+            lines = lines.to_s if Symbol === lines
+            if String === lines
+              newlines = []
+              lines.count("\n").times do
+                newlines << "\n"
+              end
+              newlines = lines.split("\n").map do |line|
+                "#{line}#{newlines.pop}"
+              end
+              if lines.starts_with?("\n")
+                newlines.insert(0, "\n")
+              end
+              newlines
+            else
+              lines
+            end
+          end
+        end
+      end
+
       UserDataFileNotFound = Class.new(StandardError)
 
       def _user_data_file(file_name, vars = {})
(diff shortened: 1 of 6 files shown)


lib/stack_master/sparkle_formation/user_data_file.rb
```diff
@@ -0,0 +1,32 @@
...
+class SparkleFormation
+  module SparkleAttribute
+    module Aws
+      class SfEruby < Erubis::Eruby
+        include Erubis::ArrayEnhancer
+
+        def add_expr(src, code, indicator)
+          case indicator
+            when '='
+              src << " #{@bufvar} << (" << code << ');'
```
taoza: Not sure what this means.

simpsora:
I'm pretty excited for this!  I haven't tested it out yet (perhaps tomorrow), but it looks very promising.

:+1: on green.

patrobinson:
Should this be a separate Gem? I think this would be useful when using plain Sparkle without stack_master.

That being said, I like this approach. 👍

lib/stack_master/sparkle_formation/user_data_file.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+      end
+
+      UserDataFileNotFound = Class.new(StandardError)
+
+      def _user_data_file(file_name, vars = {})
+        file_path = File.join(::SparkleFormation.sparkle_path, 'user_data', file_name)
+        template = File.read(file_path)
+        template_context = TemplateContext.new(vars)
+        compiled_template = SfEruby.new(template).evaluate(template_context)
```
taoza: With monkey patched `SparkleFormation::SparkleStruct` we could just `.evaluate(self)`.

spec/stack_master/sparkle_formation/user_data_file_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    klass.include(SparkleFormation::SparkleAttribute)
+    klass.include(SparkleFormation::SparkleAttribute::Aws)
+    klass.include(SparkleFormation::Utils::TypeCheckers)
+    @attr = klass.new
+    @attr._camel_keys = true
+  end
+
+  it 'reads from the user_data dir in templates' do
+    expect(File).to receive(:read).with('/templates_dir/user_data/test.erb').and_return(user_data)
+    @attr.user_data_file!('test.erb')
```
taoza: It would be good to test custom vars feature at here too.

lib/stack_master/sparkle_formation/user_data_file.rb
taoza: Ok, I think I have a rough idea what this line does. But I wonder if this can be simplified a bit like:
```
src << " #{@bufvar} << #{code};"
```

lib/stack_master/sparkle_formation/user_data_file.rb
```diff
@@ -0,0 +1,71 @@
...
+          case indicator
+            when '='
+              src << " #{@bufvar} << (" << code << ');'
+            else
+              super
+          end
+        end
+      end
+
+      class TemplateContext < AttributeStruct
```
stevehodgkiss: Why would that be better? Variables require defining methods, so this throw away class means no side effects between multiple user data evaluations in the same template/process, if we patched SF then we'd need to clean those up so I prefer this approach instead.

spec/stack_master/sparkle_formation/user_data_file_spec.rb
stevehodgkiss: Added a test for that.

lib/stack_master/sparkle_formation/user_data_file.rb
stevehodgkiss: That becomes problematic when we need to define methods for variables...

lib/stack_master/sparkle_formation/user_data_file.rb
stevehodgkiss: That breaks the test for custom vars, instead of returning the variables value it returns true.

```
       -"Fn::Base64" => {"Fn::Join"=>["", ["#!/bin/bash\n", "test_var", "\n", "yes", "\n"]]},
       +"Fn::Base64" => {"Fn::Join"=>["", ["#!/bin/bash\n", "test_var", "\n", true, "\n"]]},
```

stevehodgkiss:
Made some updates, added custom vars. I think this is ready to be merged now.

@patrobinson yeah it could be a separate gem but it could also live in SparkleFormation itself. For now I'll leave it here.

ejoubaud:
LGTM

simpsora:
:+1:, I'm pretty happy with this.  Nice job @stevehodgkiss!

Now let's get to work on command hooks and `sync_master` :-D

lib/stack_master/sparkle_formation/user_data_file.rb
taoza: Hum.. that actually make good sense.

taoza:
Belated 👍 Nice one!

PR #118, opened 2016-07-18
Use autoload and remove virtus dependency

stevehodgkiss:
By using autoload we will avoid loading dependencies when they aren't actually needed for the command that is being run.

Removes dependency on Virtus.


petervandoros:
LGTM 👍

taoza:
:+1:

PR #119, opened 2016-07-18
Support for templates backed by S3 and uploading other files as part of applying a stack

stevehodgkiss:
Carrying on from the work by @lethalpaga and @flyinbutrs.

Added a cucumber around this feature and made some other refactorings.

screen shot 2016-07-18 at 4 38 16 pm


features/apply_with_s3.feature
```diff
@@ -0,0 +1,106 @@
...
+    And a file named "parameters/myapp_web.yml" with:
+      """
+      VpcId: vpc-blah
+      """
+    And a directory named "templates"
+    And a file named "templates/user_data.sh" with:
+      """
+      #!/bin/bash
+      echo "HI"
+      """
```
stevehodgkiss: Should we looks for `files` in a different dir than `templates`? Maybe `templates/files/` or `files/` at the top level?

patrobinson: I vote for `files/`

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,66 @@
...
+          object_key = template.dup
+          object_key.prepend("#{prefix}/") if prefix
+          compiled_template_md5 = Digest::MD5.hexdigest(body).to_s
+          s3_md5 = current_objects[object_key] ? current_objects[object_key].etag.gsub("\"", '') : nil
+
+          next if compiled_template_md5 == s3_md5
+          s3_uri = "s3://#{bucket}/#{object_key}"
+          StackMaster.stdout.print "- #{File.basename(path)} => #{s3_uri} "
+
+          s3.put_object(
```
patrobinson: This won't work for large files. Should we raise an exception early? Should we document this limitation?

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        StackMaster.stdout.puts "Uploading files to S3:"
+
+        files.each do |template, file|
+          body = file.fetch(:body)
+          path = file.fetch(:path)
+          object_key = template.dup
+          object_key.prepend("#{prefix}/") if prefix
+          compiled_template_md5 = Digest::MD5.hexdigest(body).to_s
```
patrobinson: This doesn't work for multipart uploads

spec/stack_master/stack_events/presenter_spec.rb
```diff
@@ -12,7 +12,7 @@
     subject(:print_event) { described_class.print_event($stdout, event) }
 
     it "nicely presents event data" do
-      expect { print_event }.to output("\e[0;33;49m2001-01-01 02:02:02 #{Time.now.strftime('%z')} MyAwesomeQueue AWS::SQS::Queue CREATE_IN_PROGRESS Resource creation Initiated\e[0m\n").to_stdout
+      expect { print_event }.to output("\e[0;33;49m2001-01-01 02:02:02 #{time.strftime('%z')} MyAwesomeQueue AWS::SQS::Queue CREATE_IN_PROGRESS Resource creation Initiated\e[0m\n").to_stdout
```
patrobinson: I spent some time trying to find this, thank you!

spec/stack_master/test_driver/s3_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:region) { 'us-east-1' }
+
+  it 'uploads and finds files' do
+    s3_driver.upload_files(bucket: bucket,
+                           prefix: prefix,
+                           region: region,
+                           files: files)
+    file = s3_driver.find_file(bucket: bucket,
+                               object_key: [prefix, 'test-file'].compact.join('/'))
+    expect(file).to be
```
patrobinson: to be or not to be...

But seriously to be what?

patrobinson:
Should we output a diff of any non-template files uploaded?

features/apply_with_s3.feature
simpsora: Yes please.  I think `files/` makes the most sense, as the uploaded files are *not* templates.

stevehodgkiss: Agreed :)

stevehodgkiss:
@patrobinson I agree, a diff of S3 files to upload would be really useful. +1 to implementing it. I don't want to make this PR any bigger though, I'll make it a subsequent PR with more refactorings.

lib/stack_master/aws_driver/s3.rb
stevehodgkiss: What is the size limitation for this method? We could test for it and raise an exception, but Aws exceptions are already captured/handled by the CLI class and the message returned is displayed to the end user, so I'm not sure it's worth it.

patrobinson: Right the limit is 5GB, I don't think we need to worry about it...

lib/stack_master/aws_driver/s3.rb
patrobinson: As per above, probably not worth worrying about

patrobinson:
👍 on green

lib/stack_master/aws_driver/s3.rb
```diff
@@ -0,0 +1,66 @@
...
+        @region = region
+        @s3 = nil
+      end
+
+      def upload_files(bucket: nil, prefix: nil, region: nil, files: {})
+        raise StackMaster::AwsDriver::S3ConfigurationError, 'A bucket must be specified in order to use S3' unless bucket
+
+        s3 = new_s3_client(region: region)
+
+        return if files.empty?
```
petervandoros: Maybe return before creating s3 client?

petervandoros:
Very nice work! 👍

spec/stack_master/test_driver/s3_spec.rb
stevehodgkiss: To be something other than nil, essentially.

stevehodgkiss:
Release this in 0.7.0. Big thanks to @lethalpaga and @flyinbutrs for working on it.

flyinbutrs:
Re: diffing other files to be uploaded, a technique I've used in other software is to add the MD5 of the file to the metadata when you upload, which circumvents the issue of multipart uploads having mismatched etags. The only slight issue is that if you upload using a different tool that doesn't include the same metadata, you'll end up re-uploading it anyway.

PR #120, opened 2016-07-18
Support cfndsl external parameters and disabling bindings

redterror:
I'm posting this PR for feedback on this proof of concept.  It is likely not ready for merge yet.

My motivation here was to take advantage of cfndsl's support for compile-time variables.  I've always found it awkward that we have great templating power on the ruby side (and full?  nearly full? information on the CF state), but the manifestation is via CF parameters and not direct interpolation into the templates.  There are times when I want that to shorten my templates, for example by using a loop with a table of values to spread N servers across M availability zones.

What this does so far:
- Adds a pass-through key for compiler-level parameters to be specified on the stack.
- Uses the above the support 2 things for CfnDsl:
  - Call disable_binding, recommended if using ruby external_parameters.
  - Support compile-time variables (aka `external_parameters`)

If this is vaguely sane, I think next thing would be to add some reasonable default lookup strategy so we can observe the same naming convention for these parameters as we do for regular CF parameters.

Rough examples:

stack_master.yml:

```
stacks:
  us-east-1:
    nat-subnet-az-b:
      template: nat-subnet.cfndsl
      compiler_options:
        disable_binding: true
        external_parameters:
          -
            - !ruby/symbol yaml
            - cfndsl_params/us-east-1/nat_subnet_az_b.yml
```

nat-subnet.cfndsl:

```
CloudFormation do
  descr = external_parameters.fetch(:description, "Create some subnets")
  msg   = external_parameters.fetch(:fun_msg, "A default")
  Description descr
  Output(:FunMessage, msg)
end
```

cfn_params/us-east-1/nat_subnet_az_b.yml:

```
fun_msg: "WHEEEE"
```

### DIFF
--- a/lib/stack_master/template_compiler.rb
+++ b/lib/stack_master/template_compiler.rb
@@ -2,7 +2,7 @@ module StackMaster
   class TemplateCompiler
     TemplateCompilationFailed = Class.new(RuntimeError)
 
-    def self.compile(config, template_file_path, compiler_options={})
+    def self.compile(config, template_file_path, compiler_options = {})
       compiler = template_compiler_for_file(template_file_path, config)
       compiler.require_dependencies
       compiler.compile(template_file_path, compiler_options)
--- a/lib/stack_master/template_compiler.rb
+++ b/lib/stack_master/template_compiler.rb
@@ -2,10 +2,10 @@ module StackMaster
   class TemplateCompiler
     TemplateCompilationFailed = Class.new(RuntimeError)
 
-    def self.compile(config, template_file_path)
+    def self.compile(config, template_file_path, compiler_options={})
       compiler = template_compiler_for_file(template_file_path, config)
       compiler.require_dependencies
-      compiler.compile(template_file_path)
+      compiler.compile(template_file_path, compiler_options)
     rescue
       raise TemplateCompilationFailed.new("Failed to compile #{template_file_path}.")
     end
--- a/lib/stack_master/template_compilers/cfndsl.rb
+++ b/lib/stack_master/template_compilers/cfndsl.rb
@@ -4,7 +4,7 @@ module StackMaster::TemplateCompilers
       require 'cfndsl'
     end
 
-    def self.compile(template_file_path, compiler_options={})
+    def self.compile(template_file_path, compiler_options = {})
       if compiler_options["disable_binding"]
         ::CfnDsl.disable_binding
       end
--- a/lib/stack_master/template_compilers/cfndsl.rb
+++ b/lib/stack_master/template_compilers/cfndsl.rb
@@ -5,11 +5,8 @@ module StackMaster::TemplateCompilers
     end
 
     def self.compile(template_file_path, compiler_options = {})
-      if compiler_options["disable_binding"]
-        ::CfnDsl.disable_binding
-      end
-
       extras = Array(compiler_options["external_parameters"])
+      ::CfnDsl.disable_binding if !extras.empty?
       ::CfnDsl.eval_file_with_extras(template_file_path, extras).to_json
     end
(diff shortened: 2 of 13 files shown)


stevehodgkiss:
Sounds reasonable to me... Code looks good so far! 👍

lib/stack_master/template_compiler.rb
```diff
@@ -2,10 +2,10 @@ module StackMaster
   class TemplateCompiler
     TemplateCompilationFailed = Class.new(RuntimeError)
 
-    def self.compile(config, template_file_path)
+    def self.compile(config, template_file_path, compiler_options={})
```
stevehodgkiss: Whitespace around `=` would be good :)

redterror: Fixed that!

lib/stack_master/template_compilers/cfndsl.rb
```diff
@@ -4,10 +4,15 @@ def self.require_dependencies
       require 'cfndsl'
     end
 
-    def self.compile(template_file_path)
-      ::CfnDsl.eval_file_with_extras(template_file_path).to_json
+    def self.compile(template_file_path, compiler_options = {})
+      if compiler_options["disable_binding"]
+        ::CfnDsl.disable_binding
+      end
```
stevehodgkiss: Could we reset this option after compiling? It's possible to run multiple stack updates in the same process `stack_master apply production` and this config would leak to subsequent applies.

redterror: Cfndsl doesn't expose an `enable_bindings`, so to accomplish that we'd have to muck around with the module-level instance variable (see [here](https://github.com/stevenjack/cfndsl/blob/08aabe9b8d1781ae04d104e96653797580762ce8/lib/cfndsl.rb#L23)).  I get what you're saying but its a deprecated feature - perhaps the right thing is to go further and make calling `disable_binding` the default?

thebeefcake:
Rather than specifying an external_parameters although this is good as an introduction, I would like to see a similar inclusion of files using the existing inheritance/inclusion structure for yaml files. I think this will allow for greater granularity and repeatabilty.

lib/stack_master/template_compilers/cfndsl.rb
redterror: @thebeefcake - that's what the example implies, but we'd need some extra code to insulate the `disable_binding` setting in order to make it per-stack.  For CfnDsl it is a global setting with no toggle to turn it back on.

Since this is a deprecated param for CfnDsl, I think the simplest thing is really to just always force the bindings off since we're headed that way anyway.

redterror:
This also would fix #120.  I may need to extend it for a sparkleformation use case, and if so, I'll clean it up.

redterror:
The most recent commits here expand the scope of this slightly to include tweaking `sparkle_path` via the same mechanism I'm using to handle cfndsl params.  In my case I needed this + the ability to set an alternate template directory to reference a tree of sparkle unicorn turds that lives nearby but outside of my stack_master directory tree.  A simple symlink wasn't working.  I can see an argument that that is a separate concern, but it builds upon the ability to pass compiler options.

If this continues to look reasonable, then I think these are the open items:

* Support merging global/regional/stack-specific `compiler_options`.
* Pick a sensible default for cfndsl external params.  Perhaps use the same directory / names, but tack on `.cfndsl` to the extension, so `parameters/region/my_stack.yml.cfndsl`.  Thoughts / feedback?
* Additional tests.

Is there more that would be required to get this into a state where it could be merged @stevehodgkiss ?

redterror:
Out of date, future revisions to come when I get a chance.

PR #123, opened 2016-07-23
Offer to update dependent stacks

patrobinson:
This is still a little rough around the edges, but it's something that started cooking in my brain yesterday.

StackMaster encourages us to build smaller stacks. Smaller stacks are good, they make it easier to fix things when it goes wrong, like when your AutoScale Group gets wedged and you have no choice but to blow it away and re-create it. Making smaller stacks means you rely on stack_outputs to express and resolve dependencies between stacks.
But it's easy to forget about these dependencies and re-create your ASG without updating the CodeDeploy deployment group.

So here's my solution, when a stack gets updated or created check for dependent stacks and ensure their parameters are up to date. If they're not we prompt the user if they want to update that stack.


features/apply_with_dependant_stacks.feature
```diff
@@ -0,0 +1,169 @@
+Feature: Apply command
```
simpsora: Typo in filename: `dependant` -> `dependent`

This typo appears several places in this PR (sorry to be pedantic!).

lib/stack_master/commands/apply.rb
```diff
@@ -162,6 +164,13 @@ def ensure_valid_template_body_size!
         end
       end
 
+      def offer_to_run_dependant_stacks
+        @dependancies.outdated_stacks.each do |stack|
+          next unless ask?(%Q{We found a dependent stack "#{stack.stack_name}" which is now out of date because of this change.\nWould you like us to apply this stack now (y/n)?})
```
simpsora: For consistency with other prompts in `stack_master`, consider leaving out the I/We and using a more direct tone.

Compare with prompt to apply a change set:

"Apply change set (y/n)? "

simpsora:
Love this @patrobinson!  I made a couple of minor comments, but :+1: either way.

features/apply_with_dependant_stacks.feature
patrobinson: Yeah I made that spelling mistake everywhere at first, need to do a grep to find 'em all

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+    def stack_parameter(stack, key)
+      stack.parameters[key.camelize]
+    end
+
+    def output_value(key)
+      output_hash = updated_stack.outputs.select { |output_type| output_type[:output_key] == key }
+      if output_hash && ! output_hash.empty?
+        output_hash.first[:output_value]
+      else
+        raise StackOutputNotFound, "Stack exists (#{updated_stack.stack_name}), but output does not: #{key}"
```
patrobinson: Should we fail here, or just silently continue? Considering the issue is in a different stack to the one the user is "applying" I'm beginning to think we should just ignore it.

lib/stack_master/stack_dependency.rb
```diff
@@ -11,22 +11,44 @@ def outdated_stacks
...
+        any_stack_output_outdated?(parameters, dependent_stack) || any_stack_outputs_outdated?(parameters, dependent_stack)
       end
     end
 
     private
 
+    def any_stack_output_outdated?(params, stack)
+      params.any? do |_, value|
+        value['stack_output'] &&
+          value['stack_output'].gsub('_', '-') =~ %r(#{@stack_definition.stack_name}/) &&
```
simpsora: These same cleanup calls are repeated in `any_stack_outputs_outdated?`.  Will it ever be needed elsewhere?  Shall we extract it?

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def any_stack_outputs_outdated?(params, stack)
+      params.any? do |key, value|
+        value['stack_outputs'] &&
+          value['stack_outputs'].any? do |output|
+            index = value['stack_outputs'].find_index(output)
+            dependent_parameter = stack_parameter(stack, key)
+            this_output_value = dependent_parameter.split(',')[index]
+            output.gsub('_', '-') =~ %r(#{@stack_definition.stack_name}/) &&
+              output_value(output.split('/').last.camelize) != this_output_value
```
petervandoros: This can be replaced with `outdated?(stack, value['stack_output'].split('/').last)`

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+        any_stack_output_outdated?(parameters, dependent_stack) || any_stack_outputs_outdated?(parameters, dependent_stack)
+      end
+    end
+
+    private
+
+    def any_stack_output_outdated?(params, stack)
+      params.any? do |_, value|
+        value['stack_output'] &&
+          value['stack_output'].gsub('_', '-') =~ %r(#{@stack_definition.stack_name}/) &&
```
petervandoros: Can this check be extracted to a separate method?

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+          outdated?(stack, value['stack_output'].split('/').last)
+      end
+    end
+
+    def any_stack_outputs_outdated?(params, stack)
+      params.any? do |key, value|
+        value['stack_outputs'] &&
+          value['stack_outputs'].any? do |output|
```
petervandoros: Can you combine these 2 lines for indentation to better reflect the block boundaries?

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+            index = value['stack_outputs'].find_index(output)
+            dependent_parameter = stack_parameter(stack, key)
```
petervandoros: `dependent_parameter` does not change for each `output`, so consider moving it out of the `value['stack_outputs'].any? do |output|` block.

lib/stack_master/stack_dependency.rb
```diff
@@ -0,0 +1,65 @@
...
+      params.any? do |_, value|
+        value['stack_output'] &&
+          value['stack_output'].gsub('_', '-') =~ %r(#{@stack_definition.stack_name}/) &&
+          outdated?(stack, value['stack_output'].split('/').last)
+      end
+    end
+
+    def any_stack_outputs_outdated?(params, stack)
+      params.any? do |key, value|
+        value['stack_outputs'] &&
```
petervandoros: Consider checking `stack_outputs` in `any_stack_output_outdated?` to avoid loping through the same `params` twice. It will also simplify the `outdated_stacks` method.

petervandoros:
Does this solution support multi-level dependencies? E.g., If we the following dependencies:

`StackD` depends on `StackC`, which depends on `StackA`

If `StackA` has changed and you are applying `StackA`, will this propagate the changes to `StackD`? If so, how do you handle circular dependencies?

patrobinson:
@petervandoros 
> If StackA has changed and you are applying StackA, will this propagate the changes to StackD

Yes, because we're instantiating the `Apply` class again and re-issuing the command which will update and then check all dependant stacks in a recursive manner.

> If so, how do you handle circular dependencies?

Circular dependencies should be dealt with fine except if the update triggers another update of the circular dependant stack.

Consider the following examples.
Example 1:

StackB::Resource2 is dependant on StackA::Resource1
StackA::Resource2 is dependant on StackB::Resource2

StackA is updated and Resource1 is replaced
We offer to update StackB
StackB is updated and Resource2 is replaced
We offer to update StackA
StackA is updated and Resource2 is updated or replaced
End of updates, dependancies are resolved.

Example 2:

StackB::Resource1 is dependant on StackA::Resource1
StackA::Resource1 is dependant on StackB::Resource1

This will cause a loop if the update triggers a replacement. This is a rare scenario and you'd have to paint yourself into a corner to get there, but it's possible. That being said it should be visible to the operator what's going on and they can at any time hit `n` to stop the loop.

lib/stack_master/stack_dependency.rb
patrobinson: It's a slightly different case here because `value['stack_output'].split('/').last` does not return anything when we call `stack_parameter(dependent_stack, value['stack_output'].split('/').last)`.

When we use `stack_outputs` we define an input variables i.e. `subnets` and define multiple stack_outputs that form a comma separated list as the input. So:

```
Subnets:
  stack_outputs:
    - my_vpc/subnet_1
    - my_vpc/subnet_2
```

Results in an input parameter of:

```
Subnets: subnet-123abc,subnet456def
```

So we need to get that input parameter, split it and find one at the right index to compare to our output parameter.

patrobinson:
@stevehodgkiss I know from our discussion you were concerned about how long this might take to read through every single stack defined. Should we make the optional to enable behind a command line flag?

patrobinson:
Closing this for now, I think we'll end up re-using the code to instead print out a dependancy graph

PR #126, opened 2016-08-18
Upload files to S3 before creating a change set

stevehodgkiss:
The diff on update will not work if this doesn't happen before creating the change set


gstamp:
👍

petervandoros:
LGMT 👍

patrobinson:
👍

PR #127, opened 2016-08-24
Added on_failure config option to create_stack

dissonanz:


patrobinson:
Can this be added as a command line flag? I'd find that useful.

👍 either way

dissonanz:
@patrobinson: I've added the cli option. How does that look to you?

lib/stack_master/commands/apply.rb
```diff
@@ -64,7 +67,8 @@ def create_stack
           failed!("Stack creation aborted")
         end
         upload_files
-        cf.create_stack(stack_options.merge(tags: proposed_stack.aws_tags))
+        on_failure = @config.stack_defaults['on_failure'] || @options.on_failure
```
patrobinson: options is just a hash, so we can't access attributes in this way. Try `@options[:on_failure]`

dissonanz:
Needs some work.

lib/stack_master/commands/apply.rb
dissonanz: When I ran this in pry, `@options` was an Options object.
```
[1] pry(#<StackMaster::Commands::Apply>)> @options
=> <Commander::Command::Options config="test.yml", yes=true, on_failure="DELETE">
```

patrobinson: You're right when you call the CLI command it passes a `Commander::Command::Options` object, but inside our specs we use a vanilla Hash.

dissonanz:
@patrobinson I've updated the test. I couldn't reconcile using a `Commander::Command::Options` object from `cli.rb` and a hash in the test, so the test uses the commander object.

patrobinson:
LGTM cc @stevehodgkiss

dissonanz:
Any idea when this will be merged in?

lib/stack_master/commands/apply.rb
```diff
@@ -64,7 +64,8 @@ def create_stack
           failed!("Stack creation aborted")
         end
         upload_files
-        cf.create_stack(stack_options.merge(tags: proposed_stack.aws_tags))
+        on_failure = @config.stack_defaults['on_failure'] || 'ROLLBACK'
```
stevehodgkiss: I wonder if we should add this to the stack definition (which inherits `stack_defaults`) instead? That way it can be specified per stack instead of just read from the defaults.

patrobinson: To me this would be most useful when creating a fresh stack. I only want to delete the stack if it's the first time I create it because then I can't update it.

So in my use case I see no value, I see the most value in a command line flag. But that's just my use case :)

lib/stack_master/commands/apply.rb
stevehodgkiss: Should this logic be the other way around? Take cmd line args if passed in otherwise use what's in config? If we removed L15 we could say `@options.on_failure || @config.stack_defaults['on_failure'] || 'ROLLBACK'`

lib/stack_master/commands/apply.rb
dissonanz: AFAIK this only works for CREATE_STACK, so it makes sense to apply it to all stacks. Once a stack has been created successfully, it can be updated after a rollback.

stevehodgkiss:
Thanks @dissonanz!

PR #133, opened 2016-10-10
Native YAML support

### LINKED ISSUE #132: Quit converting YAML to JSON for native YAML support in CloudFormation
CloudFormation has recently supported YAML templates natively:
https://aws.amazon.com/blogs/aws/aws-cloudformation-update-yaml-cross-stack-references-simplified-substitution/

`stack_master` however still converts YAML to JSON before uploading templates, which ignores tag-based syntax such as `!Ref` and causes template validation errors. 

So it would be great if `stack_master` quits converting YAML to JSON.

skatsuta:
Resolve #132 

My implementation is a bit primitive so any comments are welcome!
(For example, maybe it'd be better if I create a new `Template` class and make it have template body and format.)

### DIFF
--- a/lib/stack_master/stack.rb
+++ b/lib/stack_master/stack.rb
@@ -9,6 +9,7 @@ module StackMaster
                 :stack_status,
                 :parameters,
                 :template_body,
+                :template_format,
                 :notification_arns,
                 :outputs,
                 :stack_policy_body,
@@ -57,6 +58,7 @@ module StackMaster
         params_hash
       end
       template_body ||= cf.get_template(stack_name: stack_name).template_body
+      template_format = identify_template_format(template_body)
       stack_policy_body ||= cf.get_stack_policy(stack_name: stack_name).stack_policy_body
       outputs = cf_stack.outputs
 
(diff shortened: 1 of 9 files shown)


stevehodgkiss:
Looks good! 👍

lib/stack_master/stack.rb
```diff
@@ -85,10 +92,17 @@ def self.generate(stack_definition, config)
           tags: stack_definition.tags,
           parameters: parameters,
           template_body: template_body,
+          template_format: template_format,
           notification_arns: stack_definition.notification_arns,
           stack_policy_body: stack_policy_body)
     end
 
+    def self.identify_template_format(template_body)
+      return :json if template_body =~ /^{/x # ignore leading whitespaces
```
jacobbednarz: do we have access to the file extension here? it might be more straight forward using that if we can.

skatsuta:
Can I get merged? Or need additional reviews?
Thanks!

stevehodgkiss:
I've released 0.8.0 with these changes. Thanks @skatsuta!

PR #136, opened 2016-10-24
WIP: add a lambda_code! helper

andrewjhumphrey:
This helper takes care of the messy details of creating and uploading the
deployment package for lambda functions that exceed what can be included in 
a stack.


andrewjhumphrey:
Closing this as the approach is less than optimal,  there should be no side effects of a template compilation.

PR #137, opened 2016-10-28
Reference remote S3 templates for stack operations

damniamgood:
My nested stack is maintained in a separate repo from stack_master. I don't want to duplicate the initialization stack code as it can go out of sync. 

Enable flag use_remote when you want to reference templates directly from S3 and skip maintaining any template locally with stack_master code


patrobinson:
Hi @damniamgood if I understand correctly you'd like to use all the StackMaster features, except manage the creation/upload of the stack template separately? Is that correct?

damniamgood:
Hi @patrobinson , 

That is precisely what I mean.

lib/stack_master/stack_definition.rb
```diff
@@ -41,6 +41,36 @@ def template_dir
...
+          target: _template_file_path,
+        ) rescue nil
+
+        if response.nil?
+          StackMaster.stderr.puts "Please double check S3 parameters in your configuration file."
+          raise "Unable to Download #{template} from S3 bucket: #{s3['bucket']}."
+        end
+
+        return _template_file_path
+      end
```
stevehodgkiss: Is there a better place for this logic? I wouldn't expect `template_file_path` to have to call S3. Maybe extract methods for the expected remote bucket/key and verify the file is there somewhere else (called from apply).

lib/stack_master/stack_definition.rb
```diff
@@ -41,6 +41,36 @@ def template_dir
...
+      if s3_configured? && s3['use_remote']
+
+        StackMaster.stdout.puts "I see use_remote flag. Templates will be referenced from S3 bucket and not locally"
+        s3_obj = ::Aws::S3::Client.new(region: s3['region'])
+        raise "Unable to Get S3 driver for downloading template" if s3_obj.nil?
+
+        # Construct file temporary file location for downloading
+        # template from S3 bucket
+        _template_dir = Dir.tmpdir()
+        _template_file_path = File.join(_template_dir, template)
```
stevehodgkiss: Do we need to prepend this with `_`?

damniamgood: Don't have to but it's a good indication in the code to show variables are temporary

lib/stack_master/stack_definition.rb
damniamgood: I will look into it

damniamgood:
Updated code based on the above comment. Still working through failed checks

Gemfile
```diff
@@ -1,4 +1,4 @@
-source 'https://rubygems.org'
+source 'http://rubygems.org'
```
jacobbednarz: We probably want to remove this to avoid accidental committing.

damniamgood: Fixed

lib/stack_master/aws_driver/s3.rb
```diff
@@ -48,6 +48,31 @@ def upload_files(bucket: nil, prefix: nil, region: nil, files: {})
...
       end
 
+
+      def download_file(bucket, prefix, region, filename)
+        raise StackMaster::AwsDriver::S3ConfigurationError, 'A bucket must be specified in order to use S3' unless bucket
+
+        s3 = new_s3_client(region: region)
+        StackMaster.stdout.puts "Downloading file from S3"
+
+        downloaded_file = File.join(Dir.tmpdir(), filename)
```
patrobinson: The parenthesis to `tmpdir` are optional and most style guides say you should exclude them.

lib/stack_master/aws_driver/s3.rb
```diff
@@ -48,6 +48,31 @@ def upload_files(bucket: nil, prefix: nil, region: nil, files: {})
...
+        response = s3.get_object(
+          {
+            bucket: bucket,
+            key: "#{prefix}/#{filename}",
+          },
+          target: downloaded_file,
+        ) rescue nil
```
patrobinson: Generally speaking we should avoid `rescue nil`, as it catches a lot of Exceptions we don't want to catch, such as a keyboard interrupt.

All s3 sdk errors are a subclass of `Aws::S3::Errors::ServiceError`, so we should instead rescue that exception. We should also do something when that exception occurs to tell the user what happened and exit cleanly.

lib/stack_master/commands/apply.rb
```diff
@@ -127,7 +127,13 @@ def update_stack
 
       def upload_files
         return unless use_s3?
-        s3.upload_files(s3_options)
+
+        # Directly use remote templates from S3 bucket
```
patrobinson: This comment really just repeats what the puts says below, and so I think it's rather redundant.

lib/stack_master/stack_definition.rb
```diff
@@ -44,7 +44,10 @@ def ==(other)
     end
 
     def template_file_path
-      File.expand_path(File.join(template_dir, template))
+      # Use Downloaded file from S3 to do further
+      # operations
+      @template_dir = s3_template_dir if s3.key?("use_remote") && s3["use_remote"] == true
```
patrobinson: It would be fairly trivial to abstract this logic into a getter method for `template_dir`:

```
def template_dir
  if s3.fetch("use_remote", false)
    s3_template_dir
  else
    @template_dir
  end
end
```

This will help with any other methods that may want to use (or may already be using) the accessor of this class.

lib/stack_master/stack_definition.rb
```diff
@@ -81,6 +84,17 @@ def s3_configured?
...
+    def s3_template_dir
+      File.dirname(
+        StackMaster.s3_driver.download_file(
+          s3['bucket'],
+          s3['prefix'],
+          s3['region'],
+          template
+          )
+      )
+    end
```
patrobinson: It would be worth memoizing this result, to avoid downloading the file everytime we need it.

patrobinson:
This PR is quite stale now so I'll be closing it. Please feel free to re-open it when it's ready for another review.

PR #139, opened 2016-11-04
Add CAPABILITY_NAMED_IAM

### LINKED ISSUE #138: InsufficientCapabilitiesException Requires capabilities : [CAPABILITY_NAMED_IAM]
My cloudformation stack require CAPABILITY_NAMED_IAM, I'm getting this error:

```
Aws::CloudFormation::Errors::InsufficientCapabilitiesException Requires capabilities : [CAPABILITY_NAMED_IAM]
```

It looks like the capability param is hardcoded https://github.com/envato/stack_master/blob/master/lib/stack_master/commands/apply.rb#L124

patrobinson:
This allows you to create named IAM resources in CloudFormation

Fixes #138

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -121,7 +121,7 @@ module StackMaster
         {
           stack_name: stack_name,
           parameters: proposed_stack.aws_parameters,
-          capabilities: ['CAPABILITY_IAM'],
+          capabilities: ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
           notification_arns: proposed_stack.notification_arns,
           stack_policy_body: proposed_stack.stack_policy_body,
           template_method => template_value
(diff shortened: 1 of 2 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -121,7 +121,7 @@ def stack_options
         {
           stack_name: stack_name,
           parameters: proposed_stack.aws_parameters,
-          capabilities: ['CAPABILITY_IAM'],
+          capabilities: ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
```
simpsora: Do we always want to blindly set these?  According to [the docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-iam-template.html#using-iam-capabilities), the intention was to make you think about what you're doing first.  Specifying these here circumvents that.

patrobinson: As does setting `CAPABILITY_IAM`

Alternatively these could be enabled with command line flags.

simpsora: Yep, I don't think we should set either one automatically, but perhaps require a flag (or a setting in `stack_master.yml`) to do so.

PR #142, opened 2016-11-08
Create a config_file! method

patrobinson:
Context
----
Embedding config files in CloudFormation templates suffers from the same pain as embedding User Data scripts. It also has a very similar solution.

I'm embedding them as an argument to my containers. While there are other ways to deploy ECS tasks than using CF, this works for now.

Changes
----

Abstract the `user_data_file!` method and allow it to be re-used for a `config_file!` method that does exactly the same thing, minus base64 encoding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ And used like this in SparkleFormation templates:
   user_data user_data_file!('app.erb', role: :worker)
 ```
 
-You can also use the `config_file!` method which evaluates templates in `templates/config/[file_name]`. It is similar to `user_data_file!` but doesn't do base64 encoding. Example:
+You can also use the `joined_file!` method which evaluates templates in `templates/config/[file_name]`. It is similar to `user_data_file!` but doesn't do base64 encoding. Example:
 
 ```
 # templates/config/someconfig.conf.erb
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -346,6 +346,27 @@ And used like this in SparkleFormation templates:
   user_data user_data_file!('app.erb', role: :worker)
 ```
 
+You can also use the `config_file!` which evaluates templates in `templates/config/[file_name]`. It is similar to `user_data_file!` but doesn't do base64 encoding. Example:
```
simpsora: ... use the `config_file!` _method_ which ...

stevehodgkiss:
Is `config_file!` the best name? What about `template_file!`?

patrobinson:
@stevehodgkiss naming things is hard. The generic class is called TemplateFile, any suggestions for that?

stevehodgkiss:
How about just `Template` and a method `template!('t.erb')`. We could also remove the `_file` from `user_data_file` or provide both options for backwards compatibility. `joined_template!`... yeah naming is hard.

patrobinson:
@stevehodgkiss I feel like we'd be over-loading the term "template" if we expose methods called `template!` or `template_file!`. Since there's already a directory of "templates", it could get a little confusing.

What about `join_from_file!` ?

stevehodgkiss:
@patrobinson `join_from_file!` sounds reasonable... Or `joined_file!`.

patrobinson:
@stevehodgkiss I went with `joined_file`, let me know if I can 🚢

Issue #143, opened 2016-11-14
accessing parameters in ruby


pickledrick:
Hi Guys,

Thanks for the Gem!

Is there a way I can access a parameter in ruby rather than use a parsed parameter in cfndsl.

For example I want to do something like

Description parameters['environment'] + 'ec2_cloudformation_stack'

Rather than an FnJoin with a Cloudformation parameter

Thanks for your help,

thebeefcake.

patrobinson:
Short answer:

Not currently

Long answer:

You cannot dereference parameters before the template is compiled by CloudFormation. DSLs like cfndsl provide a wrapper, but ultimately just compile a JSON template that's submitted to the CloudFormation API. Since StackMaster uses CloudFormation parameters, the value of that parameter is not looked up at any level until it's submitted to the API. So your DSL (whether it be Sparkleformation or cfndsl) has no capability to know the parameter value at (it's) compile time.

It might be possible to submit the parameters to cfndsl or SparkleFormation as a variable or method for dereferencing, so that it can know about them, but this functionality does not exist today.

thebeefcake:
Thanks Patrick,

I'm not sure I was clear in what I was trying to do, Is there a way to use a parameter defined in yaml as a variable in ruby/cfndsl rather than bringing it in as a cloudformation parameter.

patrobinson:
The parameters are available to stack master and Cloudformation but not the DSL (and hence no code written in the DSL) without calling out to StackMaster, and exposing a lot of implementation details to the DSL. This would be extremely fragile and likely to break with updates to StackMaster.

The only way I can see to do this would to change StackMaster to pass the parameters to the DSL library when we compile the template. This seems possible with Cfndsl.

redterror:
I started on addressing this in #120 but it still needs a little more work.

elmobp:
@redterror see #144 its complete in there

patrobinson:
Implemented in #120 and released in 1.10

PR #144, opened 2016-11-15
Adding support for cfndsl external parameters

elmobp:
To resolve this issue https://github.com/envato/stack_master/issues/143 Given I needed to access both stack_definition and config within the template compiler I have modified a few files and updated to 0.9.1 though I can revert that if required! This is a awesome piece of software I must say!

### DIFF
--- a/lib/stack_master/template_compilers/cfndsl.rb
+++ b/lib/stack_master/template_compilers/cfndsl.rb
@@ -4,10 +4,10 @@ module StackMaster::TemplateCompilers
       require 'cfndsl'
     end
 
-    def self.compile(config, template_file_path, stack_definition)
+    def self.compile(config, template_file_path, cfndsl_external_parameters)
       params = []
       params.push([:yaml, File.join(config.base_dir, 'external_parameters', 
-                  stack_definition.cfndsl_external_parameters)]) if stack_definition.cfndsl_external_parameters.length > 0
+                  cfndsl_external_parameters)]) if cfndsl_external_parameters.length > 0
       ::CfnDsl.eval_file_with_extras(template_file_path, params).to_json
     end
 
--- a/lib/stack_master/template_compilers/cfndsl.rb
+++ b/lib/stack_master/template_compilers/cfndsl.rb
@@ -6,9 +6,16 @@ module StackMaster::TemplateCompilers
 
     def self.compile(config, template_file_path, cfndsl_external_parameters)
       params = []
+      if cfndsl_external_parameters.is_a?(Array)
+        cfndsl_external_parameters.each do |cfndsl_param|
+          params.push([:yaml, File.join(config.base_dir, 'external_parameters',
+                 cfndsl_param)])      
+          end
+      else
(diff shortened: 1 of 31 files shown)


elmobp:
Apologies, fixing the tests now

gergnz:
I'm keen to see this PR merged team.

lib/stack_master/template_compilers/cfndsl.rb
```diff
@@ -4,10 +4,12 @@ def self.require_dependencies
       require 'cfndsl'
     end
 
-    def self.compile(template_file_path)
-      ::CfnDsl.eval_file_with_extras(template_file_path).to_json
+    def self.compile(config, template_file_path, stack_definition)
+      params = []
+      params.push([:yaml, File.join(config.base_dir, 'external_parameters', stack_definition.cfndsl_external_parameters)]) if stack_definition.cfndsl_external_parameters.length > 0
```
patrobinson: Can we break this line up?

elmobp: Sure, ill do that today

elmobp:
@patrobinson I not sure how what you mean by the first comment? I can move the params to stack based parms if thats what you mean

patrobinson:
@elmobp Instead of passing the entire `stack_definition` object to TemplateCompiler, we can just send the external parameters `stack_definition.cfndsl_external_parameters`

elmobp:
@patrobinson Sure! I am keen to see this move

elmobp:
@patrobinson the other reason for passing it all in was if we wanted to support an extension of another compiler the whole object is already there :) but let me know thoughts

elmobp:
@patrobinson fine with me thought we should keep it right up until compile is called? as for not using CFN param's its for this reason :) https://github.com/stevenjack/cfndsl/issues/170

gergnz:
@patrobinson I assume by Cloudformation parameters, you mean the ones you put into Cloudformation API calls. If so the reason we use yml params at compile time of cfndsl is that we use things like iteration, and all sorts of other complex logic in ruby to build the data in the final cloudformation. It would be too late to provide those as cloudformation parameters.

lib/stack_master/version.rb
```diff
@@ -1,3 +1,3 @@
 module StackMaster
-  VERSION = "0.9.0"
+  VERSION = "0.12.1"
```
stevehodgkiss: This should be `0.13.0` if we're aiming to follow semantic versioning.

elmobp: Done!

johnf:
I would also love to see this merge @patrobinson are there further changes you would still like resolved?

Disclaimer: I also work at Bulletproof, although I'm actually chasing this up for a personal project :)

patrobinson:
@johnf This needs a merge because it conflicts with `master` and it's a dirty branch, you can see a lot of commits not related to this PR here. Are you able to clean it up so I can give it another review?

patrobinson:
I think this was superseeded by #120?

PR #148, opened 2016-11-24
Add support for nested parameter resolvers for array values

petervandoros:
Add support for resolving array parameter values using parameter resolvers. This is particularly useful when building up array values from various sources.

For example, given the following parameter definition:
```
my_parameter:
  - stack_output: my-stack/output # value resolves to 'value1'
  - value2
```
The parameter value will resolve to:
```
my_parameter: 'value1,value2'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,19 @@ my_parameter:
     - value2
 ```
 
+Array parameter values can include nested parameter resolvers.
+
+For example, given the following parameter definition:
+```
+my_parameter:
+  - stack_output: my-stack/output # value resolves to 'value1'
+  - value2
+```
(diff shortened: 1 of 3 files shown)


petervandoros:
@stevehodgkiss Hmm... replacing the existing array resolver didn't cross my mind when I wrote this.

If this was to replace the existing array resolver, your example would look like:
```
my_parameter:
  - stack_output: stack/output1
  - stack_output: stack/output2
```
That doesn't look too bad even though there's a bit of duplication.

I guess people can choose which one they prefer/need.

simpsora:
Released as 0.11.0.

PR #149, opened 2016-11-29
Allow nesting user_data files

ejoubaud:
This is meant to let us DRY up our user data templates that have common parts, ala Rails templates:

```
# in dj.sh.erb

<%= render ‘common.sh.erb’ %>
# some more DJ specific instructions
```

I've only tested it in the console so far.

- [ ] Add a spec

### DIFF
--- a/lib/stack_master/sparkle_formation/template_file.rb
+++ b/lib/stack_master/sparkle_formation/template_file.rb
@@ -23,23 +23,29 @@ module StackMaster
       include ::SparkleFormation::SparkleAttribute::Aws
       include ::SparkleFormation::Utils::TypeCheckers
 
-      def self.build(vars)
+      def self.build(vars, prefix)
         ::Class.new(self).tap do |klass|
           vars.each do |key, value|
             klass.send(:define_method, key) do
               value
             end
           end
-        end.new(vars)
+
+        end.new(vars, prefix)
       end
 
-      def initialize(vars)
+      def initialize(vars, prefix)
         self._camel_keys = true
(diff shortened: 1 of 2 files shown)


patrobinson:
I was thinking about doing something like this. Looks good so far

ejoubaud:
This should be good to review now @patrobinson. cc @stevehodgkiss @gstamp too :)

ejoubaud:
As a sidenote, I stumbled on a bug where I'd get an unhelpful error when the template include an undefined variable:

```
NoMethodError: undefined method `[]' for nil:NilClass from /opt/rubies/2.1.10/lib/ruby/gems/2.1.0/gems/attribute_struct-0.3.4/lib/attribute_struct/attribute_struct.rb:300:in `method_missing'
```

Calling a missing method on our `TemplateContext` causes that error. The problem seems to be that we're overriding the `AttributeStruct` initializer with our own set of attributes.

I tried a lot of different things to solve it, always hitting a deadend. The funkiness of `AttributeStruct` always ends up getting in the way. It inherits `BasicObject` and redefines `#method_missing` and `#respond_to?` in buggy ways so inheritance doesn't work great but neither does dynamic delegation. The journey down that rabbithole was quite the cautionary tale about the worst in Ruby: spaghetti mixin design and funky method_missing-based metaprogramming. Too bad that's the road chosen by SparkleFormation, it's likely to prove limiting :(

stevehodgkiss:
Released 0.12.0 with these changes

PR #156, opened 2017-01-12
Adding support for output parameters in a different region

samuelsuarezsanchez:
There are some scenarios where using outputs from a template in a different region can be handy, for example:

* A stack might create an S3 bucket as part of `region-1`, and its name is needed in a stack which is part of `region-2`.
* When creating EC2 instances through the same stack in multiple regions, we might want to create Route53 record sets that point to the same HostedZone ID, created already by a previous template on `region-1`.
* We might want all our EC2 instances to have the same IAM role, so they all share the same set of permissions (e.g. write access to a S3 bucket). We could create a stack that deploys the IAM role as part of `region-1` and then use its ARN as a parameter for instances created in other regions.

This PR is a first attempt at supporting cross-region template output reference in parameter files.

The syntax is backwards compatible:
`[region:]stack-name/output_name`

* `region` is an optional region or region-alias (e.g. `production`) where the stack is located.
* `stack-name` is the stack name in that region.
* `output_name` is the output name of the stack.

For example:
```
# parameters/my-server.yml
# Note global is an alias of us-east-1 in this case.
instance_type: t2.micro
vpc_id:
  stack_output: my-vpc/vpc_id
public_subnets:
  stack_output: my-vpc/public_subnets
hosted_zone_id:
  stack_output: global:public-hosted-zone/zone_id
hosted_zone_name:
  stack_output: global:public-hosted-zone/zone_name
bastion_role:
  stack_output: global:bastion-host-role/name
```

### DIFF
--- a/lib/stack_master/aws_driver/cloud_formation.rb
+++ b/lib/stack_master/aws_driver/cloud_formation.rb
@@ -3,6 +3,8 @@ module StackMaster
     class CloudFormation
       extend Forwardable
 
+      attr_reader :region
+
       def set_region(region)
         @region = region
         @cf = nil
--- a/lib/stack_master/aws_driver/cloud_formation.rb
+++ b/lib/stack_master/aws_driver/cloud_formation.rb
@@ -3,11 +3,15 @@ module StackMaster
     class CloudFormation
       extend Forwardable
 
-      attr_reader :region
+      def region
+        @region ||= ENV['AWS_REGION'] || Aws.config[:region] || Aws.shared_config.region
+      end
 
-      def set_region(region)
-        @region = region
-        @cf = nil
+      def set_region(value)
+        if region != value
+          @region = value
+          @cf = nil
+        end
       end
 
       def_delegators :cf, :create_change_set,
@@ -33,7 +37,7 @@ module StackMaster
       private
 
       def cf
-        @cf ||= Aws::CloudFormation::Client.new(region: @region)
+        @cf ||= Aws::CloudFormation::Client.new(region: region)
       end
 
       def retry_with_backoff
--- a/lib/stack_master/parameter_resolvers/stack_output.rb
+++ b/lib/stack_master/parameter_resolvers/stack_output.rb
@@ -10,14 +10,14 @@ module StackMaster
         @config = config
         @stack_definition = stack_definition
         @stacks = {}
+        @output_regex = %r{(?:([^:]+):)?([^:/]+)/(.+)}
       end
 
       def resolve(value)
-        validate_value!(value)
-        stack_name, output_name = value.split('/')
-        stack = find_stack(stack_name)
+        region, stack_name, output_name = parse!(value)
+        stack = find_stack(stack_name, region)
         if stack
-          output = stack.outputs.find { |output| output.output_key == output_name.camelize }
+          output = stack.outputs.find { |o| o.output_key == output_name.camelize }
           if output
             output.output_value
           else
@@ -34,21 +34,27 @@ module StackMaster
         @cf ||= StackMaster.cloud_formation_driver
       end
 
-      def validate_value!(value)
-        if !value.is_a?(String) || !value.include?('/')
-          raise ArgumentError, 'Stack output values must be in the form of stack-name/output-name'
+      def parse!(value)
+        if !value.is_a?(String) || !(match = @output_regex.match(value))
+          raise ArgumentError, 'Stack output values must be in the form of [region:]stack-name/output_name'
         end
+
+        match.captures
       end
 
-      def find_stack(stack_name)
-        @stacks.fetch(stack_name) do
+      def find_stack(stack_name, region)
+        stack_key = stack_key(stack_name, region)
+        @stacks.fetch(stack_key) do
+          original_region = cf.region
+          cf.set_region(region) if region && original_region != region
           cf_stack = cf.describe_stacks(stack_name: stack_name).stacks.first
-          @stacks[stack_name] = cf_stack
+          cf.set_region(original_region) if region && original_region != region
+          @stacks[stack_key] = cf_stack
         end
       end
 
-      def cf
-        @cf ||= StackMaster.cloud_formation_driver
+      def stack_key(stack_name, region)
+        "#{region || 'any'}:#{stack_name}"
       end
     end
   end
--- a/lib/stack_master/parameter_resolvers/stack_output.rb
+++ b/lib/stack_master/parameter_resolvers/stack_output.rb
@@ -17,7 +17,7 @@ module StackMaster
         region, stack_name, output_name = parse!(value)
         stack = find_stack(stack_name, region)
         if stack
-          output = stack.outputs.find { |o| o.output_key == output_name.camelize }
+          output = stack.outputs.find { |stack_output| stack_output.output_key == output_name.camelize }
           if output
             output.output_value
           else
--- a/lib/stack_master/parameter_resolvers/stack_output.rb
+++ b/lib/stack_master/parameter_resolvers/stack_output.rb
@@ -39,7 +39,11 @@ module StackMaster
           raise ArgumentError, 'Stack output values must be in the form of [region:]stack-name/output_name'
         end
 
-        match.captures
+        [
+          match[:region] || cf.region,
+          match[:stack_name],
+          match[:output_name]
+        ]
       end
 
       def find_stack(stack_name, region)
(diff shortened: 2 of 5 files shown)


lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -10,14 +10,14 @@ def initialize(config, stack_definition)
...
 
       def resolve(value)
-        validate_value!(value)
-        stack_name, output_name = value.split('/')
-        stack = find_stack(stack_name)
+        region, stack_name, output_name = parse!(value)
+        stack = find_stack(stack_name, region)
         if stack
-          output = stack.outputs.find { |output| output.output_key == output_name.camelize }
+          output = stack.outputs.find { |o| o.output_key == output_name.camelize }
```
jacobbednarz: I prefer using readable variables over shorter ones so that the code flow isn't interrupted by me trying to remember what the shorter ones actually mean.

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -10,14 +10,14 @@ def initialize(config, stack_definition)
         @config = config
         @stack_definition = stack_definition
         @stacks = {}
+        @output_regex = %r{(?:([^:]+):)?([^:/]+)/(.+)}
```
jacobbednarz: I'm unsure on how I feel about using a regex here. Previously this was quite simple and would just split the string using `/`.

If we go down the route of using a regex, I'd like to see this use named captures so that we can call them using a recognisable string instead of the numeric indexes. 

@stevehodgkiss @patrobinson what are your thoughts on the syntax we should expect to use here?

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -34,21 +34,33 @@ def cf
...
-      def find_stack(stack_name)
-        @stacks.fetch(stack_name) do
+      def find_stack(stack_name, region)
+        unaliased_region = @config.unalias_region(region)
+        stack_key = stack_key(stack_name, unaliased_region)
+
+        @stacks.fetch(stack_key) do
+          original_region = cf.region
+
+          cf.set_region(unaliased_region) if region && original_region != unaliased_region
```
jacobbednarz: Is there a specific reason why we duplicated this line before and after setting the stack?

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -34,21 +34,27 @@ def cf
...
-      def find_stack(stack_name)
-        @stacks.fetch(stack_name) do
+      def find_stack(stack_name, region)
+        stack_key = stack_key(stack_name, region)
+        @stacks.fetch(stack_key) do
+          original_region = cf.region
+          cf.set_region(region) if region && original_region != region
           cf_stack = cf.describe_stacks(stack_name: stack_name).stacks.first
-          @stacks[stack_name] = cf_stack
+          cf.set_region(original_region) if region && original_region != region
```
petervandoros: Instead of changing the region of `AwsDriver::CloudFormation`, how about creating, and caching, a different instance per region instead? This way, you don't need to worry about reseting the `region` value and subsequently invaliding the underlying `Aws::CloudFormation::Client` for every stack output resolved.

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -34,21 +34,27 @@ def cf
...
+          @stacks[stack_key] = cf_stack
         end
       end
 
-      def cf
-        @cf ||= StackMaster.cloud_formation_driver
+      def stack_key(stack_name, region)
+        "#{region || 'any'}:#{stack_name}"
```
petervandoros: If the logic is changed earlier in this class, then there's no need for `|| 'any'` as `region` will never be `nil`.

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -10,14 +10,14 @@ def initialize(config, stack_definition)
...
         @stack_definition = stack_definition
         @stacks = {}
+        @output_regex = %r{(?:([^:]+):)?([^:/]+)/(.+)}
       end
 
       def resolve(value)
-        validate_value!(value)
-        stack_name, output_name = value.split('/')
-        stack = find_stack(stack_name)
+        region, stack_name, output_name = parse!(value)
```
petervandoros: It might be worth defaulting `region` to the current region to have a single path/logic resolving stack outputs and ensuring `region` will never be `nil`, simplifying all the logic.

lib/stack_master/parameter_resolvers/stack_output.rb
petervandoros: It's not duplicated, it's setting the region back to its original value.

petervandoros: Instead of changing the region of `AwsDriver::CloudFormation`, how about creating, and caching, a different instance per region instead? This way, you don't need to worry about reseting the `region` value and subsequently invaliding the underlying `Aws::CloudFormation::Client` for every stack output resolved.

petervandoros:
Also, please update the Readme documentation to include this new functionality.

samuelsuarezsanchez:
Thanks @jacobbednarz and @petervandoros for your comments. I just pushed some more changes trying to address all of them. Let me know what you think.

andrewjhumphrey:
Nice feature, nicely backwards compatible,  nicely implemented.   Basically just nice! +1 from me.

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -10,14 +10,15 @@ def initialize(config, stack_definition)
         @config = config
         @stack_definition = stack_definition
         @stacks = {}
+        @cf_drivers = {}
+        @output_regex = %r{(?:(?<region>[^:]+):)?(?<stack_name>[^:/]+)/(?<output_name>.+)}
```
jacobbednarz: ace!

petervandoros:
So it turns out the core maintainers did review this PR 😊 

👍 from me too

lib/stack_master/parameter_resolvers/stack_output.rb
```diff
@@ -34,21 +35,41 @@ def cf
...
-      def find_stack(stack_name)
-        @stacks.fetch(stack_name) do
-          cf_stack = cf.describe_stacks(stack_name: stack_name).stacks.first
-          @stacks[stack_name] = cf_stack
+      def find_stack(stack_name, region)
+        unaliased_region = @config.unalias_region(region)
+        stack_key = stack_key(stack_name, unaliased_region)
+
+        @stacks.fetch(stack_key) do
+          regional_cf = cf_for_region(region)
```
samuelsuarezsanchez: 🤦‍♂️ You're right, it should have been `unaliased_region`. Fixed in a new commit.

lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -3,9 +3,15 @@ module AwsDriver
...
-        @cf = nil
+      def region
+        @region ||= ENV['AWS_REGION'] || Aws.config[:region] || Aws.shared_config.region
+      end
+
+      def set_region(value)
+        if region != value
+          @region = value
+          @cf = nil
+        end
```
samuelsuarezsanchez: Yeah, this deserves some explanation (:warning: it's a bit long).

This solves a bunch of issues we face when region is not explicitly set. In this situation, given there's no way of getting it from the underlying `Aws::CloudFormation::Client`, we obtain it following exactly the same logic used internally by that class, so the reported region should be the same as the one used internally. 

However, in the unlikely event that AWS decides to update the logic it uses in `Client` in the future, the region used by the `Client` might differ from the region reported by us in `CloudFormation`. To avoid this, we use `CloudFormation#region` also to instantiate the internal `Client` (see below).

`set_region` is a different beast. We obliterate the underlying `Client` when someone calls this method, which is not ideal when we are using the test driver: if we call `set_region` after stubbing stacks and events, all those stubs will be lost and tests will fail.

I could have changed the tests, but my understanding is that calling `set_region` should only change it when it is a different one, thus avoiding unnecessary instantiations.

petervandoros: @samuelsuarezsanchez 💯  on the explanation. Much appreciated!

IMHO, just leave it as is!

Awesome job 👏

patrobinson:
Can we 🚢  this? cc @samuelsuarezsanchez @stevehodgkiss ?

samuelsuarezsanchez:
@patrobinson @stevehodgkiss I would think so, all requested changes have been made.

PR #160, opened 2017-02-10
Feature/region wide parameters

mikian:
Allows to define common, region-wide parameters that are shared between multiple stack (i.e. VpcId, subnets etc). Works well with base templates that defines common parameters.


redterror:
Is it worthwhile to go one step further and do a `global.yml` param, in the same vein as this?  I'm not sure if I want that or I want cross-region stack output references.  My main use case is handling of global resources, e.g. IAM stuff.

rbayerl:
Looking forward to this too - would really simplify configurations with multiple stacks :)

arecker:
+1

aglover-zendesk:
👍

johnf:
I've been using this but I think it has a problem in that anything you define in ```[region].yaml``` has to exist in every stack otherwise you get a validation error

PR #162, opened 2017-03-06
Add stackifier command

envatopoho:
I've recently been in the position where I've had to translate a cloudformation example into stack master.  It was tedious typing all the bits out, so I made a quick task to take an example file as input and turn it into a stack-master file. 

For example, the JSON found here: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-cloudfront.html

Usage is `stack_master stackify [input-file]` and will output to STDOUT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,6 +408,7 @@ stack_master events [region-or-alias] [stack-name] # Display events for a stack
 stack_master outputs [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master resources [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master status # Displays the status of each stack
+stack_master stackify [input-file] # Turns a basic cloudformation template into stackmaster syntax
 ```
 
 ## Applying updates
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -408,6 +408,7 @@ stack_master events [region-or-alias] [stack-name] # Display events for a stack
 stack_master outputs [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master resources [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master status # Displays the status of each stack
+stack_master stackify [input-file] # Turns a basic cloudformation template into stackmaster syntax
```
patrobinson: It's not stack_master syntax, it's Sparkleformation

stevehodgkiss:
I'm on the fence about whether this belongs in stack master itself, but leaning towards no. It looks useful, maybe a separate utility / ruby script would be better? Thoughts @gstamp @simpsora ?

simpsora:
@stevehodgkiss @envatopoho also on the fence.  Given the fact that it doesn't use any functionality from `stack_master`, I think either a script on the side, or perhaps a PR to SparkleFormation might be more appropriate.

gstamp:
Definitely useful but not really part of the mission for stack master. It might be useful for us to link to this sort of thing from the readme though.

PR #170, opened 2017-03-24
Add env parameter resolver

WhileLoop:


stevehodgkiss:
What's the use case for this? In your example, why wouldn't you set the database username in a parameter file?

lib/stack_master/parameter_resolvers/env.rb
```diff
@@ -0,0 +1,17 @@
...
+    class Env < Resolver
+      array_resolver
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(value)
+        ENV[value]
```
stevehodgkiss: Would it be appropriate to raise an error if the expected env var isn't present?

WhileLoop:
The use case is when using Dockerized build systems such as Codeship Pro configuration parameters are injected into the Docker container as environment variables. 

Specifically in my case, I build my application into a Docker container which I tag with the CI build id which is only available in the build system as en environment variable. I then use stack_master to create a ECS task definition which takes this tag as an input parameter.

arecker:
> I agree with @stevehodgkiss we should fail loudly and clearly if the environment variable is not present. Otherwise I'm happy with these changes

Then again, returning nothing would be consistent with other parameter resolvers.  I think the AMI lookup, for example, will just return `''` if nothing satisfies the query.  And required parameters are validated before acting on the stack.

berniedurfee-ge:
Love the idea of this feature. It would allow us to distribute CF scripts, wrapped with StackMaster, but the person executing the scripts could easily provide parameters at runtime via environment variables. We'll take a look at adding an error message to the PR, will that resolve the issue and get a merge?

patrobinson:
@berniedurfee-ge yes I'd be happy with that

Issue #174, opened 2017-04-22
Bad exit code when apply fails


rbayerl:
`stack_master` exits with a status of zero even when a stack is left in a bad state. As a result it can be difficult to notice if one stack fails as the diff for the next stack pushes the failure message off of the screen when running `apply` against all stacks.

`stack_master` should exit on failure with a non-zero status. This would allow for automated updates with Travis CI or some other tool via git hooks, and it would ensure users notice when a stack update or creation did not actually work.

PR #175, opened 2017-04-22
Exit with error if stack ends in a failed state

rbayerl:
This causes `stack_master` to raise an exception and exit if the stack creation/update fails.

Fixes https://github.com/envato/stack_master/issues/174


rbayerl:
Is there anything more needed on this? I'd be happy to refactor it to only be applied if enabled with a commandline switch if that's preferred.

Issue #176, opened 2017-04-27
Support Serverless Application Model (SAM) templates


skatsuta:
CloudFormation supports [Serverless Application Model (SAM)](https://github.com/awslabs/serverless-application-model) templates, but when we use a SAM template StackMaster returns the following error:

```sh
$ stack_master apply
...
...
No stack found
Create stack (y/n)?
Uploading files to S3:
...
Aws::CloudFormation::Errors::ValidationError CreateStack cannot be used with templates containing Transforms.
```

This is because CreateStack API does not support SAM templates.
Instead, we need to use a change set to create a SAM-based stack.
So it'd be great if StackMaster supports SAM templates.

skatsuta:
Blocker is #177.

skatsuta:
Merging both #178 and #179 will support this feature.

patrobinson:
I've released verison 0.16.0 which includes the above PRs

skatsuta:
Great! I confirmed the version works fine with SAM. Thank you for your support!

tanyajajodia:
Hi @skatsuta - Does StackMaster support the `aws cloudformation package` step as well for SAM templates? 

I get the following error when I try to create a stack directly using `stack_master apply`:
`Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [GraphQL] is invalid. 'CodeUri' is not a valid S3 Uri of the form "s3://bucket/key" with optional versionId query parameter.`

patrobinson:
@tanyajajodia sorry for the late reply. No we don't currently support that step, it must be done before StackMaster is invoked.
I'd be happy to include such functionality in the future, at present we are just using plain old SAM, but it'd be nice to directly integrate it with StackMaster.

PR #178, opened 2017-05-05
Use a change set when creating a stack

### LINKED ISSUE #177: Create a change set when creating a stack
Currently StackMaster issues `CloudFormation::Client#create_stack` when creating a stack, while it updates a stack using `CloudFormation::Client#create_change_set` and `#execute_change_set`.
However, [we can also use CreateChangeSet API when creating a stack](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-create-stacks-changesets.html).
I think it's better to use the API to create a stack because we can confirm all resources to be added by change set details.
In addition, it's required to use CreateChangeSet API to create a stack from a SAM template (So this is a blocker of #176).

I'm willing to implement this feature, but have one concern.
[`CloudFormation::Client#create_stack` can pass a stack policy with `stack_policy_body` or `stack_policy_url` parameters](http://docs.aws.amazon.com/sdkforruby/api/Aws/CloudFormation/Client.html#create_stack-instance_method), while [`CloudFormation::Client#create_change_set` cannot](http://docs.aws.amazon.com/sdkforruby/api/Aws/CloudFormation/Client.html#create_change_set-instance_method).
So I have two ideas:

1. add another command or additional option in `apply` command to create a stack using a change set
2. always create a stack using a change set and then update the stack when a stack policy exists in `stack_master.yml`

I'd like to hear your ideas and opinions to add (or against) this feature. Thank you.

skatsuta:
This resolves #177.

- Use a change set when creating a stack, instead of calling `create_stack` API
  - This allows users to review all of the resources to be created before creating a stack
- Set a stack policy after the change set is applied
  - Call `set_stack_policy` API after the stack events are complete since we cannot set a stack policy during the stack change (i.e. `*_IN_PROGRESS` state)

This is also the first part for #176.

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -12,6 +12,7 @@ module StackMaster
         @stack_definition = stack_definition
         @from_time = Time.now
         @options = options
+        @options.on_failure ||= @config.stack_defaults['on_failure']
       end
 
       def perform
@@ -63,15 +64,32 @@ module StackMaster
 
       def create_stack
         upload_files
-        options = stack_options.merge(tags: proposed_stack.aws_tags, change_set_type: 'CREATE')
-        @change_set = ChangeSet.create(options)
+        if use_change_set?
+          create_stack_by_change_set
+        else
+          create_stack_directly
+        end
+      end
+
+      def use_change_set?
+        @options.on_failure.nil?
+      end
+
+      def create_stack_by_change_set
+        @change_set = ChangeSet.create(stack_options.merge(change_set_type: 'CREATE'))
         halt!(@change_set.status_reason) if @change_set.failed?
         @change_set.display(StackMaster.stdout)
         unless ask?('Create stack (y/n)? ')
           cf.delete_stack(stack_name: stack_name)
-          halt! 'Stack creation aborted'
+          halt!('Stack creation aborted')
         end
         execute_change_set
+      end
+
+      def create_stack_directly
+        failed!('Stack creation aborted') unless ask?('Create stack (y/n)? ')
+        on_failure = @options.on_failure || 'ROLLBACK'
+        cf.create_stack(stack_options.merge(on_failure: on_failure))
       end
 
       def ask_to_cancel_stack_update
@@ -125,6 +143,7 @@ module StackMaster
         {
           stack_name: stack_name,
           parameters: proposed_stack.aws_parameters,
+          tags: proposed_stack.aws_tags,
           capabilities: ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
           role_arn: proposed_stack.role_arn,
           notification_arns: proposed_stack.notification_arns,
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -12,7 +12,6 @@ module StackMaster
         @stack_definition = stack_definition
         @from_time = Time.now
         @options = options
-        @options.on_failure ||= "ROLLBACK"
       end
 
       def perform
@@ -21,6 +20,7 @@ module StackMaster
         ensure_valid_template_body_size!
         create_or_update_stack
         tail_stack_events
+        set_stack_policy
       end
 
       private
@@ -62,12 +62,16 @@ module StackMaster
       end
 
       def create_stack
-        unless ask?('Create stack (y/n)? ')
-          failed!("Stack creation aborted")
-        end
         upload_files
-        on_failure = @config.stack_defaults['on_failure'] || @options.on_failure
-        cf.create_stack(stack_options.merge({tags: proposed_stack.aws_tags, on_failure: on_failure}))
+        options = stack_options.merge(tags: proposed_stack.aws_tags, change_set_type: 'CREATE')
+        @change_set = ChangeSet.create(options)
+        halt!(@change_set.status_reason) if @change_set.failed?
+        @change_set.display(StackMaster.stdout)
+        unless ask?('Create stack (y/n)? ')
+          cf.delete_stack(stack_name: stack_name)
+          halt! 'Stack creation aborted'
+        end
+        execute_change_set
       end
 
       def ask_to_cancel_stack_update
@@ -124,7 +128,6 @@ module StackMaster
           capabilities: ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
           role_arn: proposed_stack.role_arn,
           notification_arns: proposed_stack.notification_arns,
-          stack_policy_body: proposed_stack.stack_policy_body,
           template_method => template_value
         }
       end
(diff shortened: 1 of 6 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -62,12 +62,16 @@ def create_or_update_stack
...
+        halt!(@change_set.status_reason) if @change_set.failed?
+        @change_set.display(StackMaster.stdout)
         unless ask?('Create stack (y/n)? ')
-          failed!("Stack creation aborted")
+          cf.delete_stack(stack_name: stack_name)
+          halt! 'Stack creation aborted'
         end
-        upload_files
-        on_failure = @config.stack_defaults['on_failure'] || @options.on_failure
-        cf.create_stack(stack_options.merge({tags: proposed_stack.aws_tags, on_failure: on_failure}))
```
patrobinson: This is removing existing functionality that allows you to specify an `on_failure` option when creating a stack. This is useful when you want the stack to be deleted if it fails to create the first time.

I wonder if we can still allow this by continuing to use `create_stack` if the `on_failure` option is explicitly provided (as it will just Rollback by default) and provide some documentation to indicate you can't use that option with SAM.

lib/stack_master/commands/apply.rb
```diff
@@ -62,12 +62,16 @@ def create_or_update_stack
...
 
       def create_stack
+        upload_files
+        options = stack_options.merge(tags: proposed_stack.aws_tags, change_set_type: 'CREATE')
+        @change_set = ChangeSet.create(options)
+        halt!(@change_set.status_reason) if @change_set.failed?
+        @change_set.display(StackMaster.stdout)
         unless ask?('Create stack (y/n)? ')
-          failed!("Stack creation aborted")
+          cf.delete_stack(stack_name: stack_name)
```
patrobinson: How can you delete a stack that's never been created? Does this just delete the change set?

skatsuta: Actually when you create a change set for a new stack, an empty stack in `REVIEW_IN_PROGRESS` status is also created.
Even though you delete the change set only, the empty stack still remains.
The stack can be deleted only by `delete_stack` API.
So I think SM should delete the stack when a user decides to not create a stack.

skatsuta:
Thank you so much @patrobinson, I missed the incompatibility point.
I've updated the `create_stack` method to use either `cf.create_stack` or change set corresponding to the `on_failure` option.

lib/stack_master/commands/apply.rb
```diff
@@ -62,12 +63,33 @@ def create_or_update_stack
...
         end
-        upload_files
-        on_failure = @config.stack_defaults['on_failure'] || @options.on_failure
-        cf.create_stack(stack_options.merge({tags: proposed_stack.aws_tags, on_failure: on_failure}))
+        execute_change_set
+      end
+
+      def create_stack_directly
+        failed!('Stack creation aborted') unless ask?('Create stack (y/n)? ')
+        on_failure = @options.on_failure || 'ROLLBACK'
```
patrobinson: This doesn't need a default because `create_stack_directly` is only called when `on_failure` is set

lib/stack_master/test_driver/cloud_formation.rb
```diff
@@ -106,7 +108,7 @@ def describe_change_set(options)
       def execute_change_set(options)
         change_set_id = options.fetch(:change_set_name)
         change_set = @change_sets.fetch(change_set_id)
-        update_stack(change_set)
+        @stacks[change_set.fetch(:stack_name)].attributes = change_set
```
patrobinson: This leaves `update_stack` as dead code. Should we just change `update_stack` to not set the `stack_policy` anymore?

spec/stack_master/commands/apply_spec.rb
```diff
@@ -122,24 +139,29 @@ def apply
...
-            key: 'environment',
-            value: 'production'
-          }],
+          { key: 'environment', value: 'production' }
+        ],
         capabilities: ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
         role_arn: role_arn,
         notification_arns: [notification_arn],
-        stack_policy_body: stack_policy_body,
-        on_failure: 'ROLLBACK'
```
patrobinson: We're no longer ensuring `create_stack` is called with all these parameters when `on_failure` is set, just that `on_failure` is passed to the method. Instead of changing this spec to execute a ChangeSet, make this spec for when `on_failure` is set (and remove https://github.com/skatsuta/stack_master/blob/fa8ef933a5b7759d5be1cb3c19adbe3a7ca385af/spec/stack_master/commands/apply_spec.rb#L186-L192), then create a new spec for executing a ChangeSet

lib/stack_master/commands/apply.rb
```diff
@@ -12,7 +12,7 @@ def initialize(config, stack_definition, options = Commander::Command::Options.n
         @stack_definition = stack_definition
         @from_time = Time.now
         @options = options
-        @options.on_failure ||= "ROLLBACK"
+        @options.on_failure ||= @config.stack_defaults['on_failure']
```
patrobinson: This means if you set a stack default for `on_failure` SAM templates won't work.

Since this never worked, can we just set it to `nil` by default? I'll add some documentation in a seperate PR that this doesn't work and that's expected

skatsuta: @patrobinson Ah right, but you mean `@option.on_failure ||= nil` here?
If so, since currently `use_change_set?` returns `@option.on_failure.nil?`, do users always have to provide `--on-failure` option to call `create_stack_directly` for non-SAM templates when running `apply` command?

lib/stack_master/test_driver/cloud_formation.rb
skatsuta: I've thought it too, but if this test CFn client mocks `update_stack` API, setting a stack policy is the original behavior.
So I think it's not a reasonable change to remove [this line](https://github.com/envato/stack_master/pull/178/files#diff-e733243b5c0c71cddbe1be6d3b39a442R159) and
either leaving the `update_stack` method as dead code for future use or deleting the method totally is an option.

spec/stack_master/commands/apply_spec.rb
skatsuta: Sounds reasonable. I've added the original specs for the case when `on_failure` option is set to the below context block.

lib/stack_master/commands/apply.rb
patrobinson: Yes, that preserves the current functionality while not potentially tripping users up in the future.

lib/stack_master/test_driver/cloud_formation.rb
patrobinson: I would prefer the dead code is removed, as it adds confusion and is easily found in git history.

skatsuta: OK, removed `update_stack`.

skatsuta:
@patrobinson OK, I removed the dead code, set `on_failure` to `nil` by default.
Thank you so much for your kind review.

Issue #180, opened 2017-05-11
region-aliases and parameters used


comister:
Hi,

Just stumbled over one use-case which I think may be worth a thought.

assuming having the following stack_master.yml

```
region_aliases:
  testing: us-east-1
  production: us-east-1
  staging: ap-southeast-2
stacks:
  us-east-1:
    my-stack-name:
      template: my-cf-stack.json
  ap-southeast-2:
    my-stack-name:
      template: my-cf-stack.json
```

So with that, I thought to put parameters under `parameters/testing`, `parameters/production` and `parameters/staging` with the name `my_cf_stack.yml`.

And I was also under the assumption if I now do a `stack_master apply testing my-stack-name` it will pull the parameters only from `testing` ... but obviously due to the same region used for testing and production it will always just take 1 of the files, regardless if I use the alias for either production or testing.

Maybe there is another way to achieve that but I am not sure on that and want to point that out here.

patrobinson:
This is being worked on in #115 but it hasn't had <3 in a while. I too would like this feature, but it's not a high enough priority for me to work on at the moment.

comister:
@patrobinson Thanks for pointing that out ... I have forked the repo and started to examine stack_master ... the work got started on the feature but I somehow lost track what is still missing. To me it looks like that load_stacks is already wrongly behaving and/or was maybe the point were development stopped. Will try to do something about it ... not confident enough, yet.

comister:
I have almost understood the logic there ... but its a bit strange, is it by design that the region_alias will get replaced by the region in stack_definition, that makes not a whole lot of sense to me ... any insights into that appreciated, don't want to compromise any principles ...

comister:
Would something like this be in line with stack_master ? (it would behave as usual but if the alias itself does hold a hash it affects the way how parameters are getting resolved ... this would address my requirement, I am not 100% sure about the original requirement of adding a prefix to the stack-name)

```
region_aliases:
  testing:
    region: us-east-1
    explicit-alias-parameters: true
  production: us-east-1
  staging: ap-southeast-2
stacks:
  us-east-1:
    my-stack-name:
      template: my-cf-stack.json
  ap-southeast-2:
    my-stack-name:
      template: my-cf-stack.json```

**Comment by patrobinson** [comment]
As per the README, regions can be referred to by their alias:

```
region_aliases:
  production: us-east-1
  staging: ap-southeast-2
...
stacks:
  production:
    myapp-vpc:
[...]

PR #185, opened 2017-06-04
Environments

johnf:
A proof of concept for an implementation based on environments rather than regions.
See #180


lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -7,13 +7,24 @@ def region
         @region ||= ENV['AWS_REGION'] || Aws.config[:region] || Aws.shared_config.region
       end
 
+      def profile_name
+        @profile_name ||= ENV['AWS_PROFILE'] || Aws.config[:profile_name]
+      end
```
patrobinson: I'd like to see this in a seperate PR to have a seperate conversation there. I'm not convinced this is the right path for StackMaster.

README.md
```diff
@@ -427,17 +418,17 @@ stacks:
...
-stack_master delete [region-or-alias] [stack-name] # Delete a stack
-stack_master events [region-or-alias] [stack-name] # Display events for a stack
-stack_master outputs [region-or-alias] [stack-name] # Display outputs for a stack
-stack_master resources [region-or-alias] [stack-name] # Display outputs for a stack
+stack_master --yes apply [environment] [stack-name] # Create or update a stack non-interactively (forcing yes)
+stack_master diff [environment] [stack-name] # Display a stack tempalte and parameter diff
+stack_master delete [environment] [stack-name] # Delete a stack
+stack_master events [environment] [stack-name] # Display events for a stack
+stack_master outputs [environment] [stack-name] # Display outputs for a stack
+stack_master resources [environment] [stack-name] # Display outputs for a stack
```
patrobinson: Are we changing the core API here? Can you continue to use StackMaster in the traditional way, i.e. by just specifying a region.

johnf: there would be nothing stopping you giving environments the same names as the region they are in so yes.

However, yes this whole PR is a major breaking change. The yaml format is similar but different.

Mainly turns things around from stackmaster having regions as it's key data model entry point to environments.

lib/stack_master/aws_driver/cloud_formation.rb
johnf: I could split this out. Do you not think the whole PR is not the right direction or just this change?

Specifying the profile is completely optional. If you leave it out it will just default to the env variables.

patrobinson: I understand but I believe a Pull request should have [one concern](https://medium.com/@fagnerbrack/one-pull-request-one-concern-e84a27dfe9f1) and one concern only.

Discussions of this change are likely to confuse the other discussion that needs to happen.

README.md
patrobinson: I feel we are making a change that breaks existing functionality for the benefit of some but not all users. This doesn't feel like the right trade off to me.

I like the idea of environments, I think moving away from region aliases is a good move but this feels like the wrong approach. What about a non breaking change like:

```yaml
stack_definitions:
  myapp-vpc:
    template: vpc.rb
  myapp-elb:
    template: elb.rb
  myapp-web:
    template: web.rb

environments:
  production:
    us-east-1:
      - myapp-vpc
      - myapp-elb
      - myapp-web
    eu-central-1:
      - myapp-vpc
      - myapp-elb
      - myapp-web
  staging:
    us-west-2:
      - myapp-vpc
      - myapp-elb
      - myapp-web
```

et304383: This is better than what we have, that's for sure.  Right now I have no way to deploy the same stack  for different environments to the same region.

patrobinson:
Closing this for now, we will swing back and have a crack at implementing this when we're ready for a 2.0 release.

PR #190, opened 2017-07-24
Release the lock on aws-sdk version 2.6

gstib:
Allow any version from 2.6  to < 3 to be installed.

The hard lock is a little restrictive if you have a Gemfile and you are doing other things with the aws-sdk. It changes, new features are added and you can't use those feature with this lock. 

AWS is generally pretty good with semantic versioning. As long as we don't do a major upgrade to version 3, I would think a softer lock should work.


salamagd:
:+1: :shipit:

Issue #184, opened 2017-05-29
Support SparkleFormation compile-time parameters


simpsora:
http://www.sparkleformation.io/docs/sparkle_formation/compile-time-parameters.html

Use case: some resource types expect a particular repetitive data structure, e.g. an array of JSON objects.  There is no suitable existing variable type which supports such a thing.

Enabling compile-time parameters would expose parameters to ruby methods in the DSL, allowing the caller to create the required data structure.

This could be implemented by supporting an additional type of parameters file, whose contents are applied to compile-time parameters rather than the standard parameters.

patrobinson:
Related to #144 (which is for cfndsl)

patrobinson:
Released in 0.18.0

PR #195, opened 2017-09-08
SparkleFormation - Compile Time Parameters

gstib:
Reference #184 
Support added for Compile Time Parameter in SparkleFormation.
- Conforms to documentation [here](http://www.sparkleformation.io/docs/sparkle_formation/compile-time-parameters.html). The : prompt_when_nested option was not included as it didn't seem relevant.
- Resolved parameters get removed from AWS Cloudformation parameters if they are used as compile time parameters.


lib/stack_master/sparkle_formation/compile_time/max_size_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def check_is_valid
+          return true unless @definition[:type] == :number
+          return true unless @definition.key?(KEY)
+          invalid_values.empty?
+        end
+
+        def invalid_values
+          values = build_values(@definition, @parameter)
+          values.select {|value| value.to_f > @definition[KEY].to_f}
```
patrobinson: Nitpick, it's idiomatic in Ruby to put spaces between curly braces `{ |value| ... }`

lib/stack_master/sparkle_formation/compile_time/min_length_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def check_is_valid
+          return true unless @definition[:type] == :string
+          return true unless @definition.key?(KEY)
+          invalid_values.empty?
+        end
+
+        def invalid_values
+          values = build_values(@definition, @parameter)
+          values.select {|value| value.length < @definition[KEY].to_i}
```
patrobinson: As above.

lib/stack_master/sparkle_formation/compile_time/min_size_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def check_is_valid
+          return true unless @definition[:type] == :number
+          return true unless @definition.key?(KEY)
+          invalid_values.empty?
+        end
+
+        def invalid_values
+          values = build_values(@definition, @parameter)
+          values.select {|value| value.to_f < @definition[KEY].to_f}
```
patrobinson: As above

lib/stack_master/sparkle_formation/compile_time/number_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+        def check_is_valid
+          return true unless @definition[:type] == :number
+          invalid_values.empty?
+        end
+
+        def invalid_values
+          values = build_values(@definition, @parameter)
+          #TODO: check value is a number
+          values.reject {|value|
+            value.is_a?(Numeric) || value.is_a?(String) && value.to_s =~ /^(\d+)(\.?(\d+))?$/}
```
patrobinson: From the Github Ruby style guide:
> Prefer {...} over do...end for single-line blocks. Avoid using {...} for multi-line blocks 

See https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md

patrobinson: Further I find the logic here quite confusing... multiple and/or logic in a single line combined with regular expressions makes it hard to understand. Can this be broken down more?

lib/stack_master/sparkle_formation/compile_time/parameters_validator.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'value_validator_factory'
+
+module StackMaster
+  module SparkleFormation
+    module CompileTime
```
patrobinson: You can chain these together, like so:

```
module StackMaster::SparkleFormation::CompileTime
...
end
```

lib/stack_master/sparkle_formation/compile_time/string_validator.rb
```diff
@@ -0,0 +1,33 @@
...
+        private
+
+        def check_is_valid
+          return true unless @definition[:type] == :string
+          invalid_values.empty?
+        end
+
+        def invalid_values
+          values = build_values(@definition, @parameter)
+          values.reject {|value| value.is_a?(String)}
```
patrobinson: Nitpick: Spaces around curly braces

lib/stack_master/sparkle_formation/compile_time/value_builder.rb
```diff
@@ -0,0 +1,40 @@
...
+        def convert_strings_to_array
+          if @definition[:multiple] && @value.is_a?(String)
+            @value = @value.split(',').map(&:strip)
+          end
+        end
+
+        def convert_strings_to_numbers
+          if @definition[:type] == :number
+            @value = @value.to_f if @value.is_a?(String)
+            @value = @value.map {|item| item.is_a?(String) ? item.to_f : item} if @value.is_a?(Array)
```
patrobinson: This feels like it could be another method: `convert_array_of_string_to_numbers`

lib/stack_master/sparkle_formation/compile_time/value_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        def validate
+          @is_valid = check_is_valid
+          @error = create_error unless @is_valid
+        end
+
+        protected
+
+        def check_is_valid
+          true
```
patrobinson: The pattern for ensuring a sub-class implements this is to just raise the `NotImplemented` exception

lib/stack_master/sparkle_formation/compile_time/value_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+        end
+
+        def create_error
+
```
patrobinson: As above

lib/stack_master/sparkle_formation/compile_time/value_validator_factory.rb
```diff
@@ -0,0 +1,40 @@
...
+        VALIDATORS_TYPES = [
+            EmptyValidator,
+            StringValidator,
+            NumberValidator,
+            AllowedValuesValidator,
+            AllowedPatternValidator,
+            MaxLengthValidator,
+            MinLengthValidator,
+            MaxSizeValidator,
+            MinSizeValidator]
```
patrobinson: Nitpick: it's idiomatic to put the square bracket on a new line.

lib/stack_master/sparkle_formation/compile_time/value_validator_factory.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        def initialize(name, definition, parameter)
+          @name = name
+          @definition = definition
+          @parameter = parameter
+        end
+
+        def build
+          VALIDATORS_TYPES.map {|validator| validator.new(@name, @definition, @parameter)}
```
patrobinson: Nitpick: Spaces around curly braces

lib/stack_master/template_compilers/sparkle_formation.rb
```diff
@@ -1,18 +1,52 @@
...
+      CompileTime::ParametersValidator.new(definitions, parameters).validate
+    end
+
+    def self.create_state(definitions, parameters)
+      CompileTime::StateBuilder.new(definitions, parameters).build
+    end
 
-      JSON.pretty_generate(::SparkleFormation.compile(template_file_path))
+    def self.remove_compile_parameters(definitions, parameters)
+      definitions.each {|name, _definition| parameters.delete(name.to_s.camelize)}
```
patrobinson: Is the assumption that compile time parameters will never be used in the template? I'm not sure that's true....

lib/stack_master/validator.rb
```diff
@@ -10,8 +10,11 @@ def initialize(stack_definition, config)
     end
 
     def perform
+      parameter_hash = ParameterLoader.load(@stack_definition.parameter_files)
+      parameters = ParameterResolver.resolve(@config, @stack_definition, parameter_hash)
+
       StackMaster.stdout.print "#{@stack_definition.stack_name}: "
-      template_body = TemplateCompiler.compile(@config, @stack_definition.template_file_path, @stack_definition.compiler_options)
+      template_body = TemplateCompiler.compile(@config, @stack_definition.template_file_path, parameters, @stack_definition.compiler_options)
```
patrobinson: Time to break this into multiple lines?

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        {definition: {type: :number, allowed_pattern: '^1'}, parameter: ['1'], valid: true},
+        {definition: {type: :number, allowed_pattern: '^1'}, parameter: 2, valid: false, error: [2]},
+        {definition: {type: :number, allowed_pattern: '^1'}, parameter: '2', valid: false, error: ['2']},
+
+        {definition: {type: :number, allowed_pattern: '^1', default: '1'}, parameter: nil, valid: true},
+    ]
+
+    subject {described_class.new(name, definition, parameter).tap {|validator| validator.validate}}
+
+    scenarios.each do |scenario|
```
patrobinson: Shared examples are a much better way of implementing tests that re-use tests like this. They give a much nicer output when one fails and implement a lot of this for you. https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
```diff
@@ -0,0 +1,67 @@
+require_relative '../../../../lib/stack_master/sparkle_formation/compile_time/allowed_pattern_validator'
```
patrobinson: This should be in the `spec_helper`

patrobinson:
Also there a lot of commits on this branch, I wonder if many of them can be squashed?

lib/stack_master/template_compilers/sparkle_formation.rb
gstib: Seems odd to use them in two ways...to us. Why you use it as a compile time parameter and also as a normal parameter. 
If you only use it as a compile time parameter its going to force you include its as a normal parameter whether you use it or not

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
gstib: I'm not convinced that adds anything to readability of the test particularly if it's in another file. 
They are also not reusing the same examples between tests.

lib/stack_master/sparkle_formation/compile_time/parameters_validator.rb
gstib: Other classes are like this. Just following convention :)

lib/stack_master/sparkle_formation/compile_time/min_length_validator.rb
gstib: Might I suggest a linking tool, if a particular format is required. This is just Rubymine default formatting

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
gstib: I can pull some out..Some fail if I remove the requires. There is something odd with class loading going on in the library with specs

gstib:
You can "Squash and Merge" when the pull request is merged to master. It's an option in the dropdown :)

lib/stack_master/sparkle_formation/compile_time/min_length_validator.rb
gstib: I'll let @shivaman do the formatting.

lib/stack_master/validator.rb
gstib: Time to create a class for parameters me thinks. Wasn't going to make a structural change around existing code to much in this PR unless you really want me to

lib/stack_master/template_compilers/sparkle_formation.rb
patrobinson: A parameter can be used in multiple ways, to generate some resources and pass into user data. I don't have a solid use case though so lets see how it goes

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
patrobinson: They do use autoloading which might interfeer

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
patrobinson: You can keep shared examples in the same test file and that's what I'd encourage if they are not re-used across specs.

They provide more readable _output_ mainly, giving you what exact example failed. Without them you are left scratching your head about which specific example failed. They decorate it enough to know where to start looking.

patrobinson:
I'd prefer specific commits, "Add compile time parameter monkey patch", "add acceptance tests", "integrate with compiler" etc.

lib/stack_master/validator.rb
patrobinson: Yeah lots of changes here already. Leave it as is

lib/stack_master/sparkle_formation/compile_time/parameters_validator.rb
patrobinson: The nesting though makes this pretty aweful looking...

lib/stack_master/template_compilers/sparkle_formation.rb
shivaman: Both to and against can be argued for. I reckon in a lot of cases compile time parameters will be used to programmatically construct CFn templates, and having to include them in the resulting template will more often than not pollute the paramters file, and make it less readable. We have just been opinionated here.

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
gstib: @shivaman Why not give it a try and see which technique is easier to understand.

gstib:
Do a full squash. A lot of commits are just getting dodgy code from [here](https://github.com/sparkleformation/sfn/blob/develop/lib/sfn/command_module/template.rb) into this code base and then refactoring it into something readable and for both of us to work on it at the same time. I don't it adds a lot of value adding it to master.

lib/stack_master/sparkle_formation/compile_time/value_validator.rb
gstib: ^ @shivaman

lib/stack_master/sparkle_formation/compile_time/value_validator.rb
gstib: ^ @shivaman

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
gstib: Looks like naming is also not great in the library. Too many of the same names like "SparkleFormation" that are also in other gems

lib/stack_master/template_compilers/sparkle_formation.rb
```diff
@@ -1,18 +1,52 @@
+require_relative '../sparkle_formation/compile_time/parameters_validator'
+require_relative '../sparkle_formation/compile_time/definitions_validator'
+require_relative '../sparkle_formation/compile_time/state_builder'
+
 module StackMaster::TemplateCompilers
   class SparkleFormation
+
+    CompileTime = StackMaster::SparkleFormation::CompileTime
```
patrobinson: I've never seen this before...

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
patrobinson: naming things the same is fine, as long as you understand the repercussions. Having a top level name of SparkleFormation is a [Monkey Patch](http://culttt.com/2015/06/17/what-is-monkey-patching-in-ruby/). Defining `StackMaster::SparkleFormation` creates a seperate namespace that can be used without interfering with the SparkleFormation namespace.

Using `::SparkleFormation` ensures we are referencing the top level namespace, since by default Ruby will traverse the namespace tree (trying each parent until it finds a match).

[This](https://github.com/envato/stack_master/pull/195/files#diff-aa9e93040542f6109922463ded929141R8) I've never seen before and I'm unsure how this may impact the way libraries are loaded. Can we try without it?

lib/stack_master/template_compilers/sparkle_formation.rb
gstib: :)

spec/stack_master/sparkle_formation/compile_time/allowed_pattern_validator_spec.rb
shivaman: As discussed, shared_examples doesnt make tests much more readable in this context, and we are not reducing any lines of code. So we will leave the specs as is for the moment

lib/stack_master/sparkle_formation/compile_time/parameters_validator.rb
shivaman: Sticking with nesting modules. Consistent with other classes. Purtiness is in the eye of the beholder 😄

lib/stack_master/template_compilers/sparkle_formation.rb
```diff
@@ -1,18 +1,52 @@
+require_relative '../sparkle_formation/compile_time/parameters_validator'
+require_relative '../sparkle_formation/compile_time/definitions_validator'
+require_relative '../sparkle_formation/compile_time/state_builder'
+
```
stevehodgkiss: This can be changed to `require 'stack_master/compile_time/parameters_validation'`

gstib: Note: I noticed the use of `autoload`. This is apparently deprecated and should not be used anymore. [see](http://www.rubyinside.com/last-weeks-top-ruby-news-rails-3-1-3-autoload-deprecated-and-conferences-5650.html)

patrobinson: That's a long way off for now

patrobinson: There are 21 other `require_relative` calls, can we get rid of them too?

lib/stack_master/template_compilers/sparkle_formation.rb
```diff
@@ -1,18 +1,52 @@
...
         ::SparkleFormation.sparkle_path = File.dirname(template_file_path)
       end
+      sparkle_template = ::SparkleFormation.compile(template_file_path, :sparkle)
+      definitions = sparkle_template.parameters
+      validate_definitions(definitions)
+      validate_parameters(definitions, parameters)
+
+      sparkle_template.compile_time_parameter_setter do
+        sparkle_template.compile_state = create_state(definitions, parameters)
+        remove_compile_parameters(definitions, parameters)
```
stevehodgkiss: Not keen on how this works... relying on parameters to this method being mutated so that SM doesn't use compile time params as CF params. I wonder if it would be better to define compile time params in `stack_master.yml` under the stack definition. That way we're not mixing the 2 types of parameters. Thoughts?

```
my-stack:
  template: body.json
  compiler_options:
    parameters:
      private_subnet_cidrs:
        - 10.0.0.0/32:ap-southeast-2
        - 10.0.0.2/32:ap-southeast-1
```

gstib: You could. @patrobinson wanted to be able to also get other stacks outputs and feed them in as compile time parameters. Would it work with that?

gstib: When you use compile time parameters in sparkle it automatically generates the parameters as stack outputs. I think they wanted you to be able to feed them as inputs to other stacks. So would it follow how people expect sparkle to work?

stevehodgkiss: No it wouldn't work in that case.

It looks like the output added is just a JSON dump of the `CompiledState`, which I assume is for debugging purposes or being able to recompile a version of the template by knowing the compile time params used. If you wanted to feed anything meaningful into another stack you'd have to manually create an output.

gstib: btw if you care to have a look at how sparkle uses these things in their `sfn` utilty look no further than [here](https://github.com/sparkleformation/sfn/blob/develop/lib/sfn/command_module/template.rb)

shivaman: @stevehodgkiss Our intention is to keep it as close to how SparkleFormation works, so we dont have to create and maintain additional documentation.

patrobinson: You are maintaining the API to SparkleFormation, that doesn't prevent splitting out how parameters are handled from StackMaster's point of view.

shivaman: The mutation did feel icky when we were doing it, but we were optimising to keep the behaviour as close to SparkleFormation as possible, while leaving it open for other compilers like cfndsl.

Also the `stack_master.yml` feels like  the wrong place to implement this. These are compile time paramters and not compiler options. So it feels like the parameters yaml might be a better place to flag some parameters as compile time parameters.

gstib: I agree with @shivaman `stack_master.yml` also feels like the wrong place to me. Maybe something in the parameters.yml files like this

```
stream_name: cloudwatch-logs
shards: 1
   compile_time:
      private_subnet_cidrs:
        - 10.0.0.0/32:ap-southeast-2
        - 10.0.0.2/32:ap-southeast-1
```

I'm also thinking of how I would use them. i.e. create n-subnets in one stack. Output the subnet identifiers as a comma delimited string to be used as compile time parameters for the downstream stack.

stevehodgkiss: Sounds reasonable @gstib. That or a separate dir/file. Thoughts @patrobinson?

patrobinson: I'm happy with that suggestion

gstib:
@stevehodgkiss, @shivaman, @patrobinson can this be merged into master now?

PR #199, opened 2017-10-26
allow blank compile time parameters, as sparkle does

redterror:
Title pretty much covers it.  Blank params are legal in sparkle, so remove the restriction placed on top of them here.


gstib:
I'm not sure I agree. It's more likely a bug in Sparkle. The existing code, which compile time parameters were based on was very badly written. i would not be surprised if this was a side effect. 

The reason I think its odd it that it doesn't allow `nil` values, but it allows `blank` values. That seems somewhat inconsistent to me. I'm also trying to understand why any of the following would be desired. How would you get blank input parameters in the first place.

In the parameters file should this be valid? It will be if this change is applied.

```
compile_time_parameters:
  subnets:
  - 10.200.1.0/26
  - 
  - 10.200.1.128/26
```

redterror:
I agree it should allow `nil`, but at present it doesn't.  Big picture, my goal is to allow conditional logic to be written in sparkle templates so I can add optional features to my templates, and write code that drives those features based around non-blank CTP's.  E.g. simple stuff like:

```
if !my_param.blank?
  ... extra resources
end
```

`nil` is of course more natural, but currently prohibited by Sparkle.  For context, native cfndsl has no problem with getting `nil` via `external_parameters`.  This seemed like the shortest path til sparkle expands to allow greater control.

gstib:
Wouldn't it be better to use a boolean flag like such

- Parameter file
```
compile_time_parameters:
  feature_enabled: true
```

- Then in your sparkle template default it to be disabled like such. Then if not supplied it defaults to false.
```
SparkleFormation.new(
    :reference_vpc,
    compile_time_parameters: {
        feature_enabled: {
            type: :string,
            default: false
        }
    })
```

As a note when it comes to compile time parameters. The bulk of the logic around validation and passing through stuff to the template is in this repo. It was duplicated from here https://github.com/sparkleformation/sfn, which is not part of the core template parser. It "should" be able to allow both `nil` and `empty` if deeded desirable.

redterror:
That would be one way but booleans are not a supported class for sparkle CTP’s last I checked,  Another oversight, IMO.

All in all the documented sparkle params are fairly narrow (e.g a hash would be nice, ditto booleans), so there’s awkwardness with a bunch of options.

Allowing blanks seemed to me the least awkward and minimally different from the official sparkle docs.  I agree it would be better to have more possible classes and more descriptive validators upstream.

gstib:
I "think" more types should be possible. The validation for that is in this repo as well. It won't work for normal parameters, but "compile time parameters" it should be possible. 

Validation for what is allowed is here https://github.com/envato/stack_master/blob/master/lib/stack_master/sparkle_formation/compile_time/definitions_validator.rb. We could break from their doco and support additional types as an option

gstib:
Oh... and this is the class that does the conversion https://github.com/envato/stack_master/blob/master/lib/stack_master/sparkle_formation/compile_time/value_builder.rb

patrobinson:
Can we just include a `allow_nil` flag that defaults to false?

```
SparkleFormation.new(
    :reference_vpc,
    compile_time_parameters: {
        subnets: {
            type: :string,
            default: false
            allow_nil: true
        }
    })
```

That seems like it would satisfy both situations

gstib:
I think that would be implied with at nil value

redterror:
Hi again - I know it's been a while.  :-)  For the next round of revisions, would this be appropriate scope:

* Allow nil values, possibly _not_ by default but with a rails-style `allow_nil: true` flag.
* Allow booleans
* Maybe hashes, if I can find a way to get sparkle to play along

Earlier there was the question about blanks/nils in array CTPs.  I would suggest as a first pass stripping these ... hopefully the broader community agrees that this is expected.

Thoughts / comments?  I'd still like to get this or a similar PR merged - my team has found it incredibly helpful.

patrobinson:
> Allow nil values, possibly not by default but with a rails-style allow_nil: true flag.

If this can be implemented in this PR I'm happy to merge it

The others look like reasonable additions.

patrobinson:
Fixed in #348

PR #200, opened 2017-12-08
Fix tests not finding SecureRandom

johnf:
No idea why this doesn't fail in travis but it does on my system locally

### DIFF
--- a/lib/stack_master/test_driver/cloud_formation.rb
+++ b/lib/stack_master/test_driver/cloud_formation.rb
@@ -1,3 +1,5 @@
+require 'securerandom'
+
 module StackMaster
   module TestDriver
     class Stack


patrobinson:
What version of Ruby are you running? I just tested it with 2.4.1 and it works fine for me.

johnf:
Yeah I've very confused as well. See below we are testing 2.3.4 so weird. I'm not sure how it can possibly work without the require though. I wonder if something else is pulling it in that doesn't exist for me.

```
johnf@camilla:~$ ruby -v
ruby 2.3.1p112 (2016-04-26) [i386-linux-gnu]

johnf@camilla:~$ irb
irb(main):001:0> SecureRandom
NameError: uninitialized constant SecureRandom
Did you mean?  SecurityError
	from (irb):1
	from /usr/bin/irb:11:in `<main>'
irb(main):002:0> 
```

PR #201, opened 2017-12-08
Aws profile

johnf:
Support specifying aws profiles in the config

### DIFF
--- a/lib/stack_master/aws_driver/cloud_formation.rb
+++ b/lib/stack_master/aws_driver/cloud_formation.rb
@@ -7,9 +7,20 @@ module StackMaster
         @region ||= ENV['AWS_REGION'] || Aws.config[:region] || Aws.shared_config.region
       end
 
+      def profile_name
+        @profile_name ||= ENV['AWS_PROFILE'] || Aws.config[:profile_name]
+      end
+
       def set_region(value)
         if region != value
           @region = value
+          @cf = nil
+        end
+      end
+
+      def set_profile(value)
+        if profile_name != value
+          @profile_name = value
           @cf = nil
         end
       end
@@ -33,9 +44,16 @@ module StackMaster
       private
 
       def cf
-        @cf ||= Aws::CloudFormation::Client.new(region: region, retry_limit: 10)
-      end
+        @cf ||= begin
+          params = {
+            region: region,
+            retry_limit: 10,
+          }
+          params[:credentials] = Aws::SharedCredentials.new(profile_name: profile_name) if profile_name
 
+          Aws::CloudFormation::Client.new(params)
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


patrobinson:
I believe the AWS SDK will respect the use of the `AWS_PROFILE` environment variable, this seems like a much cleaner way to utilise profiles without writing any code for StackMaster to handle credentials.

johnf:
@patrobinson yes by default the SDK will use the environment variable. This approach is handy though one use case the China region where you will need different credentials for that stack. Means you can throw them in the stack_master config rather than having to remember to change ```AWS_PROFILE``` in your shell.

et304383:
Specifying a profile in a file that is supposed to be portable across multiple users is never a good idea.  I don't name my profile the same as you.

johnf:
I'm coming from a corporate environment where when staff start part of the setup process is setting their profile for the company to be a standard string.

lib/stack_master/aws_driver/cloud_formation.rb
```diff
@@ -33,9 +44,16 @@ def set_region(value)
...
 
       def cf
-        @cf ||= Aws::CloudFormation::Client.new(region: region, retry_limit: 10)
-      end
+        @cf ||= begin
+          params = {
+            region: region,
+            retry_limit: 10,
+          }
+          params[:credentials] = Aws::SharedCredentials.new(profile_name: profile_name) if profile_name
```
patrobinson: It seems like it would be easier to just over-write the environment variable rather than pass it down the stack.

et304383:
I worry that if you add this the next thing people will want is the ability to pass credentials via access key and secret parameters.

Then, they'll want you to have profile prompt for an MFA code if one is required.

I was on the side of wanting all these args a while back when using tools like this, but now I've learned to keep things modular and portable by just using tools like [aws-runaw](https://github.com/vancluever/aws-runas).  As long as the tool supports env vars for key, secret key, and session token (like stack_master does) then you can use aws-runas.

lib/stack_master/aws_driver/cloud_formation.rb
johnf: From memory I did it this way so that if you simply do ```stack_master apply``` different stacks can have different profiles. i.e. the Chine situation

johnf:
Happy to add tests if that helps get it landed.

patrobinson:
@et304383 This is my concern too. But I'm hoping we can tackle this problem in a 2.0 release of StackMaster where we better handle "environments" to make it easier while keeping the authentication mechanisms de-coupled. I'm not sure how we'll do that yet, open to ideas!

In the mean time I think it should be OK to keep some features, like this, which do not impact other users in the 1.x release. It's likely we'll cull them though if/when there's a better way of managing it.

patrobinson:
This PR is a bit stale. Happy to re-review when tests are added, in the mean time closing it.

PR #194, opened 2017-08-24
Delete failed change set before halt

arecker:
Fixes #193

PR #202, opened 2017-12-13
Leftover changesets

patrobinson:
Superseeds #194

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -77,7 +77,10 @@ module StackMaster
 
       def create_stack_by_change_set
         @change_set = ChangeSet.create(stack_options.merge(change_set_type: 'CREATE'))
-        halt!(@change_set.status_reason) if @change_set.failed?
+        if @change_set.failed?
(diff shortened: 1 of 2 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -76,13 +76,23 @@ def use_change_set?
...
+        begin
+          @change_set = ChangeSet.create(stack_options.merge(change_set_type: 'CREATE'))
+          if @change_set.failed?
+            ChangeSet.delete(@change_set.id)
+            halt!(@change_set.status_reason)
+          end
+
+          @change_set.display(StackMaster.stdout)
+          unless ask?('Create stack (y/n)? ')
+            cf.delete_stack(stack_name: stack_name)
```
stevehodgkiss: Why delete the stack here (when the user answers no)?

patrobinson: This is how it was before, but I think you have to delete the stack since it needs to exist to create the ChangeSet above. If you don't delete it you will not be able to `apply` it next time.

PR #203, opened 2017-12-14
Support the yaml extension for parameter files

patrobinson:


lib/stack_master/stack_definition.rb
```diff
@@ -84,16 +84,25 @@ def s3_configured?
     def additional_parameter_lookup_file_paths
       return unless additional_parameter_lookup_dirs
       additional_parameter_lookup_dirs.map do |a|
-        File.join(base_dir, 'parameters', a, "#{underscored_stack_name}.yml")
+        [
+          File.join(base_dir, 'parameters', a, "#{underscored_stack_name}.yml"),
+          File.join(base_dir, 'parameters', a, "#{underscored_stack_name}.yaml"),
```
viraptor: Does this merge both contents, or does it prefer `yml`? (and should that be documented somewhere?)

patrobinson: Good suggestion, it prefers `yml` now and it's documented and tested

lib/stack_master/stack_definition.rb
```diff
@@ -84,16 +84,25 @@ def s3_configured?
...
+      ]
     end
 
     def default_parameter_file_path
-      File.join(base_dir, 'parameters', "#{underscored_stack_name}.yml")
+      [
+        File.join(base_dir, 'parameters', "#{underscored_stack_name}.yaml"),
+        File.join(base_dir, 'parameters', "#{underscored_stack_name}.yml"),
+      ]
     end
```
stevehodgkiss: Wondering if we should rename these methods to `_paths`?

patrobinson: I've refactored the repetitive bit out. Most of the time we'd only return one path now as it's using a Glob

PR #204, opened 2017-12-20
stricter error handling

apsifly:
this is useful for ci environment to properly fail tests

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -114,7 +114,11 @@ module StackMaster
         @change_set = ChangeSet.create(stack_options)
         if @change_set.failed?
           ChangeSet.delete(@change_set.id)
-          halt!(@change_set.status_reason)
+          if @change_set.status_reason == "The submitted information didn't contain changes. Submit different information to create a change set."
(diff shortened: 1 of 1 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -114,7 +114,11 @@ def update_stack
         @change_set = ChangeSet.create(stack_options)
         if @change_set.failed?
           ChangeSet.delete(@change_set.id)
-          halt!(@change_set.status_reason)
+          if @change_set.status_reason == "The submitted information didn't contain changes. Submit different information to create a change set."
```
patrobinson: Can you use `@change_set.failed?` to determine this? Or the [ExecutionStatus](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeChangeSet.html#API_DescribeChangeSet_ResponseElements) to determine if the ChangeSet is actually executable?

This seems like it's depending too much on the API implementation. Any complexity should be wrapped up in the ChangeSet class.

patrobinson:
Due to a lack of response I'm closing this PR. Feel free to re-open it when you feel it's ready for more review.

PR #209, opened 2018-02-15
Environment Parameters

gstib:
This completes GH-170
- Raise an error if the Environment Variable is not set
- Add RSpec's
- Add happy path Cucumber test

### DIFF
--- a/features/apply_with_env_parameters.feature
+++ b/features/apply_with_env_parameters.feature
@@ -1,0 +1,49 @@
+Feature: Apply command with environment parameter
+
+  Background:
+    Given a file named "stack_master.yml" with:
+    """
+          stacks:
+            us-east-2:
+              vpc:
+                template: vpc.rb
+          """
+    And a directory named "parameters"
+    And a file named "parameters/vpc.yml" with:
+    """
+          vpc_cidr:
+            env: VPC_CIDR
+          """
+    And a directory named "templates"
+    And a file named "templates/vpc.rb" with:
+    """
+          SparkleFormation.new(:vpc) do
+
+            parameters.vpc_cidr do
+              type 'String'
+            end
+
+            resources.vpc do
+              type 'AWS::EC2::VPC'
+              properties do
+                cidr_block ref!(:vpc_cidr)
+              end
+            end
+
+          end
+          """
+
+  Scenario: Run apply and create a new stack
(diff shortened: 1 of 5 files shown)


lib/stack_master/parameter_resolvers/env.rb
```diff
@@ -0,0 +1,18 @@
...
+    class Env < Resolver
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(value)
+        environment_variable = ENV[value]
+        raise ArgumentError, "The environment variable #{value} is not set" if environment_variable.to_s.empty?
```
patrobinson: I think this should be `if environment_variable.nil?` because an empty parameter might be perfectly legitimate and `ENV` returns nil if the environment variable is _not set_

```
irb(main):007:0> ENV["FOO"]
=> nil
irb(main):008:0> exit
patrickrobinson:$ export FOO=""
patrickrobinson:$ irb
irb(main):001:0> ENV["FOO"]
=> ""
```

gstib: Feels kind of odd. 
What's the context where `nil` is undesirable but `""` string might be legitimate? Do you have an example of how you might use this?

patrobinson: `nil` is only possible if you do not set the Environment variable. Any empty string is only possible if you _explicitly_ set it to an empty string. In that case we should trust the user knows what they are doing and did this explicitly.

gstib: @stevehodgkiss what are you're thoughts on this. Should it be allowed?

stevehodgkiss: I think an empty string is a valid input we should support. We could check `ENV.key?(value)` or `environment_variable.nil?` as @patrobinson mentioned.

features/apply_with_env_parameters.feature
```diff
@@ -0,0 +1,49 @@
...
+
+            resources.vpc do
+              type 'AWS::EC2::VPC'
+              properties do
+                cidr_block ref!(:vpc_cidr)
+              end
+            end
+
+          end
+          """
```
stevehodgkiss: Indentation is off...

PR #210, opened 2018-02-21
Make Change Set output more visible

patrobinson:
## Context

At the moment the change set output isn't easily distinguishable from the JSON and Parameter diff.

```
Stack diff: No changes
Parameters diff:
 ---
 LambdaAlias: current
 LambdaFunction: my_lambda_function
-MaxBatchSize: '1'
+MaxBatchSize: '10'
 StreamArn: arn:aws:kinesis:us-east-1:012345678910:stream/foo
Proposed change set:
Modify AWS::Lambda::EventSourceMapping EventSource
- Properties.BatchSize. Triggered by: DirectModification(Dynamic).
- Properties.BatchSize. Triggered by: ParameterReference.MaxBatchSize.
Apply change set (y/n)?
```

This has tripped people up in the past where they have not noticed the change set and even deleted production databases.
I believe the Change Set is the single most important thing people should look at and therefore the most visible.

## Change

Make the output more visible by adding a line break and border to the change set.

Before:
screen shot 2018-02-21 at 3 07 05 pm

After:
screen shot 2018-02-21 at 3 19 18 pm

### DIFF
--- a/lib/stack_master/change_set.rb
+++ b/lib/stack_master/change_set.rb
@@ -42,11 +42,11 @@ module StackMaster
     end
 
     def display(io)
-      io.puts <<~EOL
+      io.puts <<-EOL
 
-      ==========
-      Proposed change set:
-      EOL
+==========
+Proposed change set:
+EOL
       @response.changes.each do |change|
         display_resource_change(io, change.resource_change)
       end
--- a/lib/stack_master/change_set.rb
+++ b/lib/stack_master/change_set.rb
@@ -42,9 +42,11 @@ module StackMaster
     end
 
     def display(io)
-      io.puts "
+      io.puts <<~EOL
+
       ==========
-      Proposed change set:"
+      Proposed change set:
+      EOL
       @response.changes.each do |change|
         display_resource_change(io, change.resource_change)
       end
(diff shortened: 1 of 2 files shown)


lib/stack_master/change_set.rb
```diff
@@ -42,7 +42,9 @@ def initialize(describe_change_set_response)
     end
 
     def display(io)
-      io.puts "Proposed change set:"
+      io.puts "
```
lukerobins: nitpick. Is this line break needed?

patrobinson: I think it'll look better. Here's what it would look like without (new images added to PR description)

screen shot 2018-02-21 at 3 05 42 pm

lib/stack_master/change_set.rb
```diff
@@ -42,7 +42,11 @@ def initialize(describe_change_set_response)
     end
 
     def display(io)
-      io.puts "Proposed change set:"
+      io.puts <<-EOL
+
+==========
```
simpsora: Extreme nitpick: can this separator be extended to be longer, either the same length as `Proposed change set:` or perhaps even longer, to enforce the visual separation?

patrobinson: Better?
screen shot 2018-02-21 at 3 19 18 pm

simpsora: ❤️

PR #211, opened 2018-02-21
parameter_store parameter resolver.

andrewjhumphrey:
Resolves parameters from the AWS SSM parameter store service as alternative
to the dotgpg malarky. (which is great for large teams but is overkill for single
person/small team use)

This is the inital cut, just uses default AWS managed encryption keys.

Format is pretty simple
`parameter.yml`:
```
foo:
  parameter_store: ssm_name
```
Fetches the SecureString parameter ssm_name from the AWS SSM store using the prevailing
aws credentials.

DONE:  
1. Add ability to specify KMS key_id to use to decrypt 
       (to make it useful for storing things like database passwords)  (Can't get_parameter doesn't let you specify the key to use, it's stored with the parameter at encryption time)

1. Add ability to specify version to use to decrypt (Can't retrieve a particular version so of limited utility so not going to do)
1. Make it work with parameter arrays (Not required, 1 parameter 1 value and it can be a comma separated list if required)

TODO:
Nothing

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,19 @@ db_password:
   secret: db_password
 ```
 
+### Parameter Store
+An alternative to the secrets store, uses the AWS SSM Parameter store to protect
+secrets.   Expects a parameter of either `String` or `SecureString` type to be present in the
+same region as the stack. You can store the parameter using a command like this
+`aws-vault exec <whatever> -- aws ssm put-parameter --region <region> --name <parameter name> --value <secret> --type (String|SecureString)`
+When doing so make sure you don't accidentally store the secret in your `.bash_history` and
+you will likely want to set the parameter to NoEcho in your template.
+
+```yaml
+db_password:
+  parameter_store: ssm_parameter_name
+```
+
 ### Security Group
 
 Looks up a security group by name and returns the ARN.
--- a/features/apply_with_parameter_store_parameters.feature
+++ b/features/apply_with_parameter_store_parameters.feature
@@ -12,9 +12,9 @@ Feature: Apply command with parameter_store parameter
     And a file named "parameters/vpc.yml" with:
       """
       vpc_cidr:
-        parameter_store: CUCUMBER_TEST_VPC_CIDR
+        parameter_store: "/cucumber-test-vpc-cidr"
       """
-    And a SSM parameter named "CUCUMBER_TEST_VPC_CIDR" with value "10.0.0.0/16" in region "us-east-2"
+    And a SSM parameter named "/cucumber-test-vpc-cidr" with value "10.0.0.0/16" in region "us-east-2"
     And a directory named "templates"
     And a file named "templates/vpc.rb" with:
       """
--- a/features/apply_with_parameter_store_parameters.feature
+++ b/features/apply_with_parameter_store_parameters.feature
@@ -1,0 +1,47 @@
+Feature: Apply command with parameter_store parameter
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us-east-2:
+          vpc:
+            template: vpc.rb
+      """
+    And a directory named "parameters"
+    And a file named "parameters/vpc.yml" with:
+      """
+      vpc_cidr:
+        parameter_store: CUCUMBER_TEST_VPC_CIDR
+      """
+    And a SSM parameter named "CUCUMBER_TEST_VPC_CIDR" with value "10.0.0.0/16" in region "us-east-2"
+    And a directory named "templates"
+    And a file named "templates/vpc.rb" with:
+      """
+      SparkleFormation.new(:vpc) do
+
+        parameters.vpc_cidr do
+          type 'String'
+        end
+
+        resources.vpc do
+          type 'AWS::EC2::VPC'
+          properties do
+            cidr_block ref!(:vpc_cidr)
+          end
+        end
+
+      end
+      """
+
+  Scenario: Run apply and create a new stack
+    Given I stub the following stack events:
+      | stack_id | event_id | stack_name | logical_resource_id | resource_status | resource_type              | timestamp           |
+      | 1        | 1        | vpc        | Vpc                 | CREATE_COMPLETE | AWS::EC2::VPC              | 2020-10-29 00:00:00 |
+      | 1        | 1        | vpc        | vpc                 | CREATE_COMPLETE | AWS::CloudFormation::Stack | 2020-10-29 00:00:00 |
+    When I run `stack_master apply us-east-2 vpc --trace`
+    And the output should contain all of these lines:
+      | +---                  |
+      | +VpcCidr: 10.0.0.0/16 |
+    And the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} vpc AWS::CloudFormation::Stack CREATE_COMPLETE/
+    Then the exit status should be 0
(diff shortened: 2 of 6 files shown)


lib/stack_master/parameter_resolvers/parameter_store.rb
```diff
@@ -0,0 +1,24 @@
...
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(value)
+        resp = ssm.get_parameter(
+          name: value,
+          with_decryption: true
+        )
+        resp.parameter.value
```
petervandoros: Do you know how array parameter values are handled by StackMaster? Does a `StringList` parameter type need to be transformed into an `Array` or will it be handled just fine by StackMaster?

spec/stack_master/parameter_resolvers/parameter_store_spec.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.describe StackMaster::ParameterResolvers::ParameterStore do
+
+  describe '#resolve' do
+
+    let(:config) { double(base_dir: '/base') }
+    let(:stack_definition) { double(stack_name: 'mystack', region: 'us-east-1') }
```
patrobinson: this should probably be an `instance_double`, which validates any methods called on it actually exist on an instance of the StackDefinition class

spec/stack_master/parameter_resolvers/parameter_store_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        Aws.config[:ssm] = {
+          stub_responses: {
+            get_parameter: 
+              Aws::SSM::Errors::ParameterNotFound.new(unknown_parameter_name, "Parameter #{unknown_parameter_name} not found")
+          }
+        }
+      end
+      it 'should raise and error' do
+        expect { resolver.resolve(unknown_parameter_name) }
+            .to raise_error(Aws::SSM::Errors::ParameterNotFound, "Parameter #{unknown_parameter_name} not found")
```
patrobinson: I think we should catch and re-raise this exception with a different exception class we define. That's at least what all the other parameter resolvers do.

features/step_definitions/parameter_store_steps.rb
```diff
@@ -0,0 +1,15 @@
+Given(/^(?:a|the) SSM parameter(?: named)? "([^"]*)" with value "([^"]*)" in region "([^"]*)"$/) do |parameter_name, parameter_value, parameter_region|
+  File.open("/tmp/humpy-log", 'a') { |file| file.write("In Here name: #{parameter_name} value: #{parameter_value} \n") }
```
patrobinson: I'm assuming this was in here for debugging? Can it be removed?

features/apply_with_parameter_store_parameters.feature
```diff
@@ -0,0 +1,47 @@
...
+          vpc:
+            template: vpc.rb
+      """
+    And a directory named "parameters"
+    And a file named "parameters/vpc.yml" with:
+      """
+      vpc_cidr:
+        parameter_store: CUCUMBER_TEST_VPC_CIDR
+      """
+    And a SSM parameter named "CUCUMBER_TEST_VPC_CIDR" with value "10.0.0.0/16" in region "us-east-2"
```
stevehodgkiss: Parameters wouldn't usually be upper case. Maybe change this to something like `/cucumber-test-vpc-cidr` or `/myapp/prod/something-api-key`?

features/step_definitions/parameter_store_steps.rb
andrewjhumphrey: Yep, has been 17501ef  Was debugging why everything appeared to be stubbed, before I discovered `lib/stack_master/testing.rb`

features/apply_with_parameter_store_parameters.feature
andrewjhumphrey: Shall do, the uppercase was a hangover where I was actually creating the parameter in SSM parameter store (before I realised that all aws calls are stubbed in test mode) and I didn't want to trip over real parameters.

lib/stack_master/parameter_resolvers/parameter_store.rb
andrewjhumphrey: No idea,  I'll add that to the todo list to figure out.

spec/stack_master/parameter_resolvers/parameter_store_spec.rb
andrewjhumphrey: Addressed in 8e729dd following pattern found in parameter_resolvers/secret

features/apply_with_parameter_store_parameters.feature
andrewjhumphrey: Done: acd7f6f

lib/stack_master/parameter_resolvers/parameter_store.rb
andrewjhumphrey: I don't actually thing it's an issue here,  there is only ever one parameter with the same name, so you can only get one value, so iterating and fetching isn't sensible.  (which is how the other array_parameters work) 
That said if you need to you could store "1,2,3,4,5" as the value for the parameter and it will just work as all stackmaster does is concatenate with comma separators (at least according to my reading of https://github.com/envato/stack_master/blob/master/lib/stack_master/resolver_array.rb#L9-L13)

spec/stack_master/parameter_resolvers/parameter_store_spec.rb
andrewjhumphrey: *nod* Probably but dunno it's worth rejigging everything in this PR, it's done this way in:
- `parameter_resolvers/latest_ami_by_tags_spec`
- `parameter_resolvers/latest_ami_spec`
- `stack_master/parameter_resolvers/env_spec`
...etc
and I'd prefer to address this in a separate PR if we want to change the approach across a bunch of resolvers...

lib/stack_master/parameter_resolvers/parameter_store.rb
```diff
@@ -1,16 +1,23 @@
...
 
       def resolve(value)
-        resp = ssm.get_parameter(
-          name: value,
-          with_decryption: true
-        )
+        begin
+          resp = ssm.get_parameter(
+            name: value,
+            with_decryption: true
```
patrobinson: Does this fail if the parameter is not encrypted?

README.md
```diff
@@ -240,6 +240,19 @@ db_password:
   secret: db_password
 ```
 
+### Parameter Store
+An alternative to the secrets store, uses the AWS SSM Parameter store to protect
+secrets.   Expects a parameter of either `String` or `SecureString` type to be present in the
+same region as the stack. You can store the parameter using a command like this
+`aws-vault exec <whatever> -- aws ssm put-parameter --region <region> --name <parameter name> --value <secret> --type (String|SecureString)`
```
patrobinson: I don't think we should reference `aws-vault` here, as many people use different authentication methods.

patrobinson: This needs some more white space to avoid being included on the same line

lib/stack_master/parameter_resolvers/parameter_store.rb
```diff
@@ -0,0 +1,31 @@
+module StackMaster
+  module ParameterResolvers
+    class ParameterStore < Resolver
+
+      SecretNotFound = Class.new(StandardError)
```
patrobinson: I think `ParameterNotFound` would be more appropriate here

andrewjhumphrey: SecretNotFound was already used, but sure naming is hard :-)

lib/stack_master/parameter_resolvers/parameter_store.rb
andrewjhumphrey: Nope, it's ignored for anything except a SecureString

lib/stack_master/parameter_resolvers/parameter_store.rb
andrewjhumphrey: Updated in 4d9ae15

lib/stack_master/parameter_resolvers/parameter_store.rb
petervandoros: Yeah, that's my understanding also.

PR #216, opened 2018-02-28
Allow loading of sparkle packs

patrobinson:
SparkleFormation supports the concept of a [SparklePack](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html) which can be packaged as a RubyGem and required inside the template that utilises it. This allows re-usable dynamics to be shared very simply.

Unfortunately the implementation is [fundamentally](https://github.com/sparkleformation/sparkle_formation/pull/239) [broken](https://github.com/sparkleformation/sparkle_formation/pull/240). Such that requiring a sparkle pack inside a template like so (which is what the documentation suggests):

```
require 'vpc_sparkle_pack'
pack = ::SparkleFormation::SparklePack.new(:name => 'vpc_sparkle_pack')
SparkleFormation.new(:util_vpc, sparkle: pack) do
...
end
```

results in a race condition, dependent on the file load order which is not deterministic.

This bug occurs when SparkleFormation tries to load a SparklePack while it's still loading `templates/`. Therefore the easiest workaround is to pre-load the SparklePack from StackMaster.

I would've liked to have fixed this in SparkleFormation, but after spending 3 days on the problem I feel I am no closer to a solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,6 +481,16 @@ stacks:
           vpc-sparkle-pack: vpc-sparkle-pack
 ```
 
+The template can then simply load a dynamic from the sparkle pack like so:
+
+```ruby
+SparkleFormation.new(:my_stack_with_dynamic) do
+   dynamic!(:sparkle_pack_dynamic)
+end
+```
+
+Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
+
 ## Commands
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ stacks:
     my-stack:
       template: my-stack-with-dynamic.rb
       compiler_options:
-        sparkle_pack:
+        sparkle_packs:
           vpc-sparkle-pack: vpc-sparkle-pack
 ```
 
--- a/README.md
+++ b/README.md
@@ -467,6 +467,20 @@ stacks:
       template: my-stack.rb
 ```
 
+### Loading SparklePacks
+
+[SparklePacks](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html) can be pre-loaded using compiler options. This requires the name of a rubygem to `require` followed by the name of the SparklePack, which is usually the same name as the Gem.
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: my-stack-with-dynamic.rb
+      compiler_options:
+        sparkle_pack:
+          vpc-sparkle-pack: vpc-sparkle-pack
+```
+
 ## Commands
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -478,7 +478,7 @@ stacks:
       template: my-stack-with-dynamic.rb
       compiler_options:
         sparkle_packs:
-          vpc-sparkle-pack: vpc-sparkle-pack
+          - vpc-sparkle-pack
 ```
 
 The template can then simply load a dynamic from the sparkle pack like so:
--- a/lib/stack_master/stack.rb
+++ b/lib/stack_master/stack.rb
@@ -65,7 +65,7 @@ module StackMaster
       parameter_hash = ParameterLoader.load(stack_definition.parameter_files)
       template_parameters = ParameterResolver.resolve(config, stack_definition, parameter_hash[:template_parameters])
       compile_time_parameters = ParameterResolver.resolve(config, stack_definition, parameter_hash[:compile_time_parameters])
-      template_body = TemplateCompiler.compile(config, stack_definition.template_file_path, compile_time_parameters,stack_definition.compiler_options)
+      template_body = TemplateCompiler.compile(config, stack_definition.template_file_path, compile_time_parameters, stack_definition.compiler_options)
       template_format = TemplateUtils.identify_template_format(template_body)
       stack_policy_body = if stack_definition.stack_policy_file_path
                             File.read(stack_definition.stack_policy_file_path)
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -467,6 +467,20 @@ stacks:
...
+
+[SparklePacks](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html) can be pre-loaded using compiler options. This requires the name of a rubygem to `require` followed by the name of the SparklePack, which is usually the same name as the Gem.
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: my-stack-with-dynamic.rb
+      compiler_options:
+        sparkle_pack:
```
gstib: Is this meant to be `sparkle_packs`

README.md
```diff
@@ -467,6 +467,20 @@ stacks:
       template: my-stack.rb
 ```
 
+### Loading SparklePacks
+
+[SparklePacks](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html) can be pre-loaded using compiler options. This requires the name of a rubygem to `require` followed by the name of the SparklePack, which is usually the same name as the Gem.
```
gstib: Might be worth stating that we are deviating from SparkleFormation prescribed technique. 
i.e. You wont be doing this

```
require 'my_sparkle_pack'

sfn = SparkleFormation.new(:my_template) do
   dynamic!(:my_dynamic)
end
sfn.sparkle.add_sparkle(custom_pack)
```
It's just
```
SparkleFormation.new(:my_template) do
   dynamic!(:my_dynamic)
end
```

README.md
```diff
@@ -467,6 +467,20 @@ stacks:
...
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: my-stack-with-dynamic.rb
+      compiler_options:
+        sparkle_pack:
+          vpc-sparkle-pack: vpc-sparkle-pack
```
gstib: This line is a little confusing as well. I can see it's `gemname: packname` but aren't they the same thing...
Can you get rid of the `packname`

README.md
patrobinson: Fixed in `b593862`

README.md
patrobinson: Added in 978dfea

README.md
patrobinson: I don't know, but SparkleFormation supports it so I'm inclined to maintain that interface unless there's some compelling reason not to.

gstib: When you traditionally loaded like this

```
require 'my_sparkle_pack'

custom_pack = SparkleFormation::SparklePack.new(:name => 'my_sparkle_pack')
sfn = SparkleFormation.new(:my_template) do
   dynamic!(:my_dynamic)
end
sfn.sparkle.add_sparkle(custom_pack)
``` 
They actually needed to be the same to work

patrobinson: From what I understand it's entirely possible to load it like this:

```ruby
require 'my_sparkle_pack'

custom_pack = SparkleFormation::SparklePack.new(:name => 'wow_dude_a_sparkle_pack')
```

Where `my_sparkle_pack` registers itself like so:

```ruby
SparkleFormation::SparklePack.register!('wow_dude_a_sparkle_pack')
```

gstib: @patrobinson that's undocumented, you have to physically look at the code to see that. Why would you make it hard to consume by naming the pack different to the gem? Seems an odd thing to do to me.

If you look at the how they tell you to search for sparkle-pack gems [here]( http://www.sparkleformation.io/docs/sfn/sparkle-packs.html) and look at gems from [here](https://rubygems.org/search?query=sparkle-pack)

No-one does that has implemented a pack has done that. They've all followed the doco from [here](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html#distribution)  

https://github.com/reverseskate/sparkle-pack-cfn-init-chef/blob/master/lib/sparkle-pack-cfn-init-chef.rb
https://github.com/timurb/sparkle-pack-shell-helpers/blob/master/lib/sparkle-pack-shell-helpers.rb
https://github.com/reverseskate/sparkle-pack-aws-availability-zones/blob/master/lib/sparkle-pack-aws-availability-zones.rb
https://github.com/reverseskate/sparkle-pack-aws-amis/blob/develop/lib/sparkle-pack-aws-amis.rb
https://github.com/carnivore-rb/sparkle-pack-jackal-cfn/blob/master/lib/sparkle-pack-jackal-cfn.rb
https://github.com/timurb/sparkle-pack-rails-helpers/blob/master/lib/sparkle-pack-rails-helpers.rb
https://github.com/reverseskate/sparkle-pack-cfn-init-chef/blob/master/lib/sparkle-pack-cfn-init-chef.rb

I wouldn't cater for an undocumented edge case personally. I'd make it easy to use and understand.

README.md
```diff
@@ -467,6 +467,30 @@ stacks:
...
+[SparklePacks](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html) can be pre-loaded using compiler options. This requires the name of a rubygem to `require` followed by the name of the SparklePack, which is usually the same name as the Gem.
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: my-stack-with-dynamic.rb
+      compiler_options:
+        sparkle_packs:
+          vpc-sparkle-pack: vpc-sparkle-pack
```
gstib: You might need to change `sparkle_packs` to a list of sparkle packs to preserve order. I'm not sure if this structure will guarantee it. Ordering looks important if you want to override things. See [here](http://www.sparkleformation.io/docs/sparkle_formation/sparkle-packs.html#usage)

README.md
patrobinson: It's a supported API but whatever, done in 00e8993

patrobinson:
> If Sparkleformation fixes it at some time in the future, will we revert this?

Yes, I think it's best to follow the SparkleFormation documented golden path _where possible_

PR #217, opened 2018-03-18
Fix S3 url for China region

toshyak:
There is slightly different AWS S3 url for China region.


patrobinson:
Fixed in release 1.3.1

PR #220, opened 2018-03-25
Add 1password secret lookup using the 1password cli (`op`)

denmat:
This adds 1password lookups using the 1password cli `op`.

You declare your parameters like so:

```
parameters/database.yml
database_password:
  one_password:
    title: production database
    vault: Shared
    type: password
```

Currently we support two types of secrets, `password`s and `secureNote`s.

You must sign in before being able to resolve secrets:
`eval $(op signin <url> <email> <SECRET>)`
see [here](https://support.1password.com/command-line-getting-started/) for more detail

### DIFF
--- a/lib/stack_master/parameter_resolvers/one_password.rb
+++ b/lib/stack_master/parameter_resolvers/one_password.rb
@@ -1,0 +1,48 @@
+module StackMaster
+  module ParameterResolvers
+    class OnePassword < Resolver
+      OnePasswordNotFound = Class.new(StandardError)
+
+      array_resolver
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(params={})
+        raise RuntimeError, "1password requires the `OP_SESSION_<name>` to be set" if ENV.keys.grep(/OP_SESSION_\w+$/).empty?    
+        get_items(params)
+      end
+
+      private
+
+      def op_get_item(item, vault)
+        begin
+          %x(op get item --vault='#{vault}' '#{item}')
+        rescue Errno::ENOENT
+          raise RuntimeError, "The op cli needs to be installed in the PATH"
+        rescue => exception
+          raise RuntimeError, exception
+        end
+      end
+
+      def get_password(title, vault)
+        JSON.parse(op_get_item(title, vault)).dig('details', 'fields').select { |k,v| k.key('password') }[0].values_at('value').first
+      end
+      
+      def get_secure_note(title, vault)
+        JSON.parse(op_get_item(title, vault))['details'].values_at('notesPlain').first
+      end
+
+      def get_items(params)
+        case params['type']
+        when 'password'
+          return get_password(params['title'], params['vault'])
+        when 'secureNote'
+          return get_secure_note(params['title'], params['vault'])
+        end
+      end
+    end
+  end
+end
--- a/lib/stack_master/parameter_resolvers/one_password.rb
+++ b/lib/stack_master/parameter_resolvers/one_password.rb
@@ -35,6 +35,10 @@ module StackMaster
         raise OnePasswordInvalidResponse, "Failed to parse JSON returned, #{item}: #{exception}"
       end
 
+      def is_login_item?(data)
+        data.details.password.nil?
+      end
+
       def op_get_item(item, vault)
         validate_op_installed?
         item = %x(op get item --vault='#{vault}' '#{item}' 2>&1)
@@ -46,9 +50,16 @@ module StackMaster
       end
 
       def get_password(title, vault)
-        create_struct(title, vault).details.fields[1].value
+        # There are two types of password that can be returned.
+        # One is attached to a Login item in 1Password
+        # the other is to a Password item.
+        if is_login_item?(create_struct(title, vault))
+          create_struct(title, vault).details.fields[1].value
+        else
+          create_struct(title, vault).details.password
+        end
       end
-      
+
       def get_secure_note(title, vault)
         create_struct(title, vault).details.notesPlain
       end
--- a/lib/stack_master/parameter_resolvers/one_password.rb
+++ b/lib/stack_master/parameter_resolvers/one_password.rb
@@ -21,8 +21,8 @@ module StackMaster
 
       def validate_op_installed?
         %x(op --version)
-        rescue Errno::ENOENT => exception
-          raise OnePasswordBinaryNotFound, "The op cli needs to be installed and in the PATH, #{exception}"
+      rescue Errno::ENOENT => exception
+        raise OnePasswordBinaryNotFound, "The op cli needs to be installed and in the PATH, #{exception}"
       end
 
       def validate_response?(item)
@@ -30,8 +30,8 @@ module StackMaster
           raise OnePasswordNotFound, "Failed to return item from 1password, #{i['error']}"
         end
         JSON.parse(item)
-        rescue JSON::ParserError => exception
-          raise OnePasswordInvalidResponse, "Failed to parse JSON returned, #{item}: #{exception}"
+      rescue JSON::ParserError => exception
+        raise OnePasswordInvalidResponse, "Failed to parse JSON returned, #{item}: #{exception}"
       end
 
       def op_get_item(item, vault)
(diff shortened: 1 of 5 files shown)


lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      private
+
+      def op_get_item(item, vault)
+        begin
+          %x(op get item --vault='#{vault}' '#{item}')
+        rescue Errno::ENOENT
+          raise RuntimeError, "The op cli needs to be installed in the PATH"
+        rescue => exception
```
andrewjhumphrey: Does this throw a sensible error message when and item isn't present in the vault?

denmat: does now :)

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      private
+
+      def op_get_item(item, vault)
+        begin
+          item = %x(op get item --vault='#{vault}' '#{item}' 2>&1)
+          return item if JSON.parse(item)
+        rescue Errno::ENOENT
+          raise RuntimeError, "The op cli needs to be installed in the PATH"
```
patrobinson: _and_ in the PATH?

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,49 @@
...
+        rescue => exception
+          raise RuntimeError, "Failed to return item from 1password, #{item}"
```
patrobinson: Should we print the exception too?

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,49 @@
...
+        raise RuntimeError, "1password requires the `OP_SESSION_<name>` to be set" if ENV.keys.grep(/OP_SESSION_\w+$/).empty?    
+        get_items(params)
+      end
+
+      private
+
+      def op_get_item(item, vault)
+        begin
+          item = %x(op get item --vault='#{vault}' '#{item}' 2>&1)
+          return item if JSON.parse(item)
```
patrobinson: I think this would swallow any potential JSON parsing errors...

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,49 @@
...
+        rescue Errno::ENOENT
+          raise RuntimeError, "The op cli needs to be installed in the PATH"
+        rescue => exception
+          raise RuntimeError, "Failed to return item from 1password, #{item}"
+        end
+      end
+
+      def get_password(title, vault)
+        JSON.parse(op_get_item(title, vault))['details']['fields'].select { |k,v| k.key('password') }[0].values_at('value').first
```
patrobinson: I think this could be wrapped up a bit better to make it easier to grok and deal with. Should the results from `op_get_item` be an OpenStruct or a Class? This would also enable us to test against multiple versions of the `op` client version if we need to.

denmat: I toyed with `OpenStruct` previously. I didn't realise I could add it as a output to `JSON.parse`! This will change my world in so many ways.

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -17,12 +17,25 @@ def resolve(params={})
 
       private
 
+      def validate_op_installed?
+        %x(op --version)
+        rescue Errno::ENOENT => exception
+          raise RuntimeError, "The op cli needs to be installed and in the PATH, #{exception}"
```
patrobinson: Should these two lines be indented one level less? Ruby isn't white space sensitive, so it doesn't matter but for style ya know

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -17,12 +17,25 @@ def resolve(params={})
...
+      end
+
+      def parseable_json?(item)
+        JSON.parse(item)
+        rescue TypeError => exception
+          raise "item returned is not valid JSON: #{item}"
+        rescue JSON::ParserError => exception
+          raise "Failed to parse JSON returned, #{item}"
```
patrobinson: Same indentation question for these

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -41,12 +41,16 @@ def op_get_item(item, vault)
         end
       end
 
+      def create_struct(title, vault)
+        JSON.parse(op_get_item(title, vault), object_class: OpenStruct)
+      end
+
       def get_password(title, vault)
-        JSON.parse(op_get_item(title, vault))['details']['fields'].select { |k,v| k.key('password') }[0].values_at('value').first
+        create_struct(title, vault).details.fields[1].value
```
patrobinson: hot diggidy

lib/stack_master/parameter_resolvers/one_password.rb
patrobinson: Should this also print the exception?

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      array_resolver
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(params={})
+        raise RuntimeError, "1password requires the `OP_SESSION_<name>` to be set" if ENV.keys.grep(/OP_SESSION_\w+$/).empty?
```
simpsora: Should this have a more useful error message?  Maybe mention that you have to sign in first?

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,66 @@
...
+        raise RuntimeError, "1password requires the `OP_SESSION_<name>` to be set" if ENV.keys.grep(/OP_SESSION_\w+$/).empty?    
+        get_items(params)
+      end
+
+      private
+
+      def validate_op_installed?
+        %x(op --version)
+        rescue Errno::ENOENT => exception
+          raise RuntimeError, "The op cli needs to be installed and in the PATH, #{exception}"
```
simpsora: Indentation question applies to these two lines as well.

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -0,0 +1,66 @@
...
+        rescue TypeError => exception
+          raise "item returned is not valid JSON: #{item}"
+        rescue JSON::ParserError => exception
+          raise "Failed to parse JSON returned, #{item}"
+      end
+
+      def op_get_item(item, vault)
+        validate_op_installed?
+        begin
+          item = %x(op get item --vault='#{vault}' '#{item}' 2>&1)
```
simpsora: What's the reasoning behind redirecting `stderr` to `stdout`?  I doubt `stderr` will be outputting JSON, which is a requirement the line below.  If there was any output on `stderr` this would probably fail.

Better to let it write to `stderr`?  Or send to `/dev/null` if you don't care about it?

denmat: No, it captures the error message from `op`, which we now handle a bit better

lib/stack_master/parameter_resolvers/one_password.rb
denmat: updated error handling now

lib/stack_master/parameter_resolvers/one_password.rb
```diff
@@ -20,25 +22,22 @@ def resolve(params={})
       def validate_op_installed?
         %x(op --version)
         rescue Errno::ENOENT => exception
-          raise RuntimeError, "The op cli needs to be installed and in the PATH, #{exception}"
+          raise OnePasswordBinaryNotFound, "The op cli needs to be installed and in the PATH, #{exception}"
       end
 
-      def parseable_json?(item)
+      def validate_response?(item)
+        item.match(/\[LOG\].+(?<error>\(.+)$/)  do |i|
```
patrobinson: What happens if the output is something other than an `ERROR`?

denmat: It will be picked up as invalid JSON. An error appears to have a `[LOG]` prefix if it gets an error response from `op` so it should be picked up here, otherwise it should have invalid JSON and be picked up there. 

```
Please enter your password for user denmat:
[LOG] 2018/04/16 15:00:18 (ERROR) 401: Authentication required.
```

PR #221, opened 2018-03-27
support multiple aliases for a single region

rbayerl:
Related to https://github.com/envato/stack_master/pull/115 and https://github.com/envato/stack_master/issues/180 I have hit an issue where only the last region alias defined is actually honored. This simple patch allows for multiple aliases in a single region, but does not set any precedence for parameter resolution. E.g. consider the following:
```
region_aliases:
  foo: us-east-1
  bar: us-east-1

stacks:
  foo:
    foo-stack:
      template: foo.rb
  bar:
    bar-stack:
      template: bar.rb
```
Parameters for both stacks will look in `parameters/foo` and `parameters/bar`. Since one cannot have duplicate stack names in a region IMO this is not a huge issue. This patch will at least allow the option for logical `stack_master.yml` layout and ordering for parameter files.

### DIFF
--- a/lib/stack_master/config.rb
+++ b/lib/stack_master/config.rb
@@ -96,7 +96,8 @@ module StackMaster
 
     def resolve_region_aliases(stacks)
       stacks.inject({}) do |hash, (region, attributes)|
-        hash[unalias_region(region)] = attributes
+        hash[unalias_region(region)] ||= {}
+        hash[unalias_region(region)].merge!(attributes)
         hash
       end
     end


rbayerl:
Using the above example config here's what comes from `irb`:
```
irb(main):001:0> require 'stack_master'
=> true
irb(main):002:0> config = StackMaster::Config.load!('stack_master.yml')
=> #<StackMaster::Config:0x007fa261042440 @config={"template_dir"=>"./templates", "region_aliases"=>{"foo"=>"us-east-1", "bar"=>"us-east-1"}, "stacks"=>{"foo"=>{"foo-stack"=>{"template"=>"foo.rb"}}, "bar"=>{"bar-stack"=>{"template"=>"bar.rb"}}}}, @base_dir="/Users/rbayerl/foobar", @template_dir="./templates", @stack_defaults={}, @region_aliases={"foo"=>"us-east-1", "bar"=>"us-east-1"}, @region_to_aliases={"us-east-1"=>["foo", "bar"]}, @region_defaults={}, @stacks=[#<StackMaster::StackDefinition:0x007fa261029af8 @additional_parameter_lookup_dirs=["foo", "bar"], @compiler_options={}, @notification_arns=[], @s3={}, @files=[], @template="bar.rb", @region="us-east-1", @stack_name="bar-stack", @base_dir="/Users/rbayerl/foobar", @template_dir="./templates">], @template_compilers={:rb=>:sparkle_formation, :json=>:json, :yml=>:yaml, :yaml=>:yaml}>
irb(main):003:0> config.stacks[0].additional_parameter_lookup_dirs
=> ["foo", "bar"]
```
Since the parameter lookup directories just creates an array of all aliases these will all be included for all stacks in the region (regardless of the alias where the stack is defined). So parameters for the `foo` stack will look in `['parameters/foo.yml', 'parameters/us-east-1/foo.yml', 'parameters/foo/foo.yml', 'parameters/bar/foo.yml']`. Hope that's at least a little more clear.

PR #222, opened 2018-03-29
aws-sdk v3 limited to only required gems

rbayerl:
Move to `aws-sdk` v3 and minimize footprint to only use required gems.

Saw this in the Vision document and figured I would help out. I tested all of the cli commands and everything is still working AFAICT. LMK if I can help with anything else or if there is more to do here.


stack_master.gemspec
```diff
@@ -28,7 +28,11 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "timecop"
   spec.add_dependency "ruby-progressbar"
   spec.add_dependency "commander"
-  spec.add_dependency "aws-sdk", "~> 2.6"
+  spec.add_dependency "aws-sdk-cloudformation"
+  spec.add_dependency "aws-sdk-ec2"
+  spec.add_dependency "aws-sdk-s3"
+  spec.add_dependency "aws-sdk-sns"
+  spec.add_dependency "aws-sdk-ssm"
```
patrobinson: Can you please pin these to a major version number?

rbayerl:
Added version pins.

Issue #140, opened 2016-11-04
Add compile command


simpsora:
It would be helpful to see the compiled template.  You could use the `apply` command and say no at the prompt, but if a stack has unmet dependencies (e.g. an unresolved stack output), it will show an error message and exit.

Adding a command which compiled the template and output it for examination would be helpful.

patrobinson:
Fixed in #245

PR #225, opened 2018-04-20
add print command to show compiled template

rbayerl:
Fixes https://github.com/envato/stack_master/issues/140

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,6 +508,7 @@ stack_master events [region-or-alias] [stack-name] # Display events for a stack
 stack_master outputs [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master resources [region-or-alias] [stack-name] # Display outputs for a stack
 stack_master status # Displays the status of each stack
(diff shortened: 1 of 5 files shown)


rbayerl:
`sfn print` deals with compile time parameters by prompting the user. `stack_master` has parameter files to avoid this behavior so I've opted to pull in compile time parameters from the parameter file.

`stack_master print` will work without any parameter file as long as there are no compile time parameters. If there are compile time parameters in the template that do not have default values then those values will need to be populated in the parameter file for the stack.

rbayerl:
@patrobinson anything else needed here?

patrobinson:
@viraptor this implements the same functionality you were working on. Can we use this PR instead?

viraptor:
I don't mind either version.
One difference I spotted is that this one always outputs json, but my version does {json,rb} -> json and yaml -> yaml.

patrobinson:
Oh right we already merged that

@rbayerl sorry this was super-seeded by #245 which I'll release a new version for shortly.

rbayerl:
:+1: works for me

Issue #92, opened 2016-03-31
New parameter resolvers


stevehodgkiss:
It would be useful to have parameter resolvers for the following:

- EC2 KeyPair ID by name
- SSL Certificate ID by name
- IP address by EIP allocation ID

PR #227, opened 2018-04-24
add resolver for acm certificates

rbayerl:
Partially addresses https://github.com/envato/stack_master/issues/92

@stevehodgkiss I'm not sure what you're looking for wrt EC2 KeyPairs or EIP allocation IDs.

EC2 key pairs are specified by key name in Cloudformation so I'm not certain how useful a resolver would be or what it would resolve.

EIP associations are done on the EIP resource (not the instance itself) so not sure what to resolve here either. Is the use case that given `x` allocation ID just return the public IP? What's the use case?

Happy to add more with clarification 😄

### DIFF
--- a/lib/stack_master/parameter_resolvers/acm_certificate.rb
+++ b/lib/stack_master/parameter_resolvers/acm_certificate.rb
@@ -1,0 +1,37 @@
+module StackMaster
+  module ParameterResolvers
+    class AcmCertificate < Resolver
+      CertificateNotFound = Class.new(StandardError)
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(domain_name)
+        cert_arn = find_cert_arn_by_domain_name(domain_name)
+        raise CertificateNotFound, "Could not find certificate #{domain_name} in #{@stack_definition.region}" unless cert_arn
+        cert_arn
+      end
+
+      private
+
+      def all_certs
+        certs = []
+        next_token = nil
+        client = Aws::ACM::Client.new(region: @stack_definition.region)
+        loop do
+          resp = client.list_certificates(certificate_statuses: ['ISSUED'], next_token: next_token)
+          certs << resp.certificate_summary_list
+          next_token = resp.next_token
+          break if next_token.nil?
+        end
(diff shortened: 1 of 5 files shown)


lib/stack_master/parameter_resolvers/acm_certificate.rb
```diff
@@ -0,0 +1,32 @@
...
+    class AcmCertificate < Resolver
+      CertificateNotFound = Class.new(StandardError)
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(value)
+        cert_arn = all_certs.map { |c| c.certificate_arn if c.domain_name == value }.compact.first
```
stevehodgkiss: Maybe a extract method with this logic (`find_cert_arn_by_domain_name`?)

rbayerl: :+1:

lib/stack_master/parameter_resolvers/acm_certificate.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      private
+
+      def all_certs
+        certs = []
+        next_token = nil
+        client = Aws::ACM::Client.new(region: @stack_definition.region)
+        loop do
+          resp = client.list_certificates(next_token: next_token)
```
patrobinson: Should we filter for only `ISSUED` certificates?

rbayerl: Good call. Done.

patrobinson:
@rbayerl I'm OK with dropping support for 2.1

rbayerl:
Might need to clear the cache on Travis.

rbayerl:
Could be I was wrong and the test just fails intermittently. Seems you didn't hit this issue with your recent PR. I can drop the last commit if needed.

patrobinson:
@rbayerl I'm still ok with it, please resolve the merge conflict and I'll merge

rbayerl:
Dropped that commit anyway (unintentionally).

rbayerl:
@patrobinson anything else needed here?

PR #229, opened 2018-05-14
Windows build scripts

patrobinson:
This creates some scripts to build a Windows Gem for StackMaster.

Envato employees can use our internal CI system to build a Gem when releasing a new version.


Dockerfile.windows.ci
```diff
@@ -0,0 +1,38 @@
+# Temp Core Image
+FROM microsoft/windowsservercore AS core
+
+ENV RUBY_VERSION 2.2.4
```
petervandoros: Why this old version of Ruby?

patrobinson: ETOOHARD basket. Maybe another day, I don't think it should make much of a difference.

Issue #232, opened 2018-05-16
Error when stack is in state 'REVIEW_IN_PROGRESS'


denmat:
If you abort the `stack_master apply` or don't have permissions to delete the cloudformation stack, your stack is left in `REVIEW_IN_PROGRESS`.

Subsequent applies will fail with a large trace that doesn't lead you to what the problem is.

```
bundler: failed to load command: stack_master (/Users/denmat/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/bin/stack_master)
NoMethodError: undefined method `fetch' for false:FalseClass
  /Users/denmat/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/bundler/gems/stack_master-48c829d8a17c/lib/stack_master/stack.rb:20:in `template_default_parameters'
```

PR #233, opened 2018-05-16
Adds a check for REVIEW_IN_PROGRESS on CF stack.

denmat:
relates to issue https://github.com/envato/stack_master/issues/232

Adds a check on `REVIEW_IN_PROGRESS` status in the Cloudformation stack which can be left if that state by a user or someone else.

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -46,11 +46,20 @@ module StackMaster
         !stack.nil?
       end
 
+      def stack_in_review?
+        return false if stack.nil?
+        if stack.stack_status == "REVIEW_IN_PROGRESS"
+          StackMaster.stderr.puts "Stack currently exists and is in #{stack.stack_status}"
+          failed! "You will need to delete the stack (#{stack.stack_name}) before continuing"
+        end
+      end
+
       def use_s3?
         !@s3_config.empty?
       end
 
       def diff_stacks
+        stack_in_review?
         StackDiffer.new(proposed_stack, stack).output_diff
       end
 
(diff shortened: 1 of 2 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -46,11 +46,19 @@ def stack_exists?
         !stack.nil?
       end
 
+      def abort_if_review_in_progress
+        if stack_exists? && stack.stack_status == "REVIEW_IN_PROGRESS"
```
patrobinson: I don't think `REVIEW_IN_PROGRESS` is necessarily a bad state? Do we get in this state if we propose a change and select `n` to applying it?

The main problem we encounter is an empty stack

patrobinson: We do cleanup if you ctrl c now, but can't if you there are no permissions. I'm just wondering that instead of checking for the status, should we check for an empty stack instead?

denmat: hmm here is the result of `stack.inspect` when we get this issue.

```
#<StackMaster::Stack:0x00007fd95d50b070 @region="us-east-1", @stack_name="mistack", @stack_id="arn:aws:cloudformation:us-east-1:123456789:stack/mistack-database/6c029cb0-5901-11e8-a5fa-500c2854b699", @parameters={}, @template_body="", @template_format=:yaml, @outputs=[], @role_arn=nil, @notification_arns=[], @stack_policy_body=nil, @stack_status="REVIEW_IN_PROGRESS">
```
do you think we can test on template_body?

PR #243, opened 2018-06-15
Update Config.rb to raise error on badly formatted yaml

denmat:
With badly formatted yaml I get this error message, which usually takes some time for me to remember what the problem is.

```
stack_master validate
error: (<unknown>): mapping values are not allowed in this context at line 19 column 19. Use --trace to view backtrace
```

Now, this helps me remember in a 🔦 :
```
stack_master validate
error: Unable to parse <path-to-stack_master.yml>: (<unknown>): mapping values are not allowed in this context at line 19 column 19. Use --trace to view backtrace
```

### DIFF
--- a/lib/stack_master/config.rb
+++ b/lib/stack_master/config.rb
@@ -3,11 +3,15 @@ require 'active_support/core_ext/object/deep_dup'
 
 module StackMaster
   class Config
+    ConfigParseError = Class.new(StandardError)
+
     def self.load!(config_file = 'stack_master.yml')
       resolved_config_file = search_up_and_chdir(config_file)
       config = YAML.load(File.read(resolved_config_file))
       base_dir = File.dirname(File.expand_path(resolved_config_file))
       new(config, base_dir)
+    rescue Psych::SyntaxError => error
+      raise ConfigParseError, "Unable to parse stack_master.yml: #{error}"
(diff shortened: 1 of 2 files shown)


lib/stack_master/config.rb
```diff
@@ -3,11 +3,15 @@
...
   class Config
+    ConfigParseError = Class.new(StandardError)
+
     def self.load!(config_file = 'stack_master.yml')
       resolved_config_file = search_up_and_chdir(config_file)
       config = YAML.load(File.read(resolved_config_file))
       base_dir = File.dirname(File.expand_path(resolved_config_file))
       new(config, base_dir)
+    rescue Psych::SyntaxError => error
+      raise ConfigParseError, "Unable to parse stack_master.yml: #{error}"
```
patrobinson: I'd love it if we could create our own wrapper for yaml parser and catch this, it's a pain in parameter files and templates too

viraptor: stack_master.yml is hardcoded in the message. Should it be `#{config_file}` instead?

denmat: Yes, it should `stack_master.yml` anyways, but using the `resolved_config_file` would probably help more too. See d62f9ba

PR #245, opened 2018-07-11
Add lint and compile commands

viraptor:
Make it easy to use cfn-lint on the templated stacks. The templated
stack is written to a temporary file which cfn-lint is run on. Since
the error messages contain the path to the json node, it's not that
important to preserve the rendered json with matching lines.

Unfortunately there's no special error code for success/failure. We can
just forward the original output.

Having cfn-lint is not a requirement for stack_master and the missing
binary is handled without exploding.

Also add a `compile` command which outputs the processed stack
template for use in other tools.

### DIFF
--- a/lib/stack_master/commands/lint.rb
+++ b/lib/stack_master/commands/lint.rb
@@ -19,6 +19,7 @@ module StackMaster
 
         Tempfile.open(['stack', ".#{proposed_stack.template_format}"]) do |f|
           f.write(proposed_stack.template_body)
+          f.flush
           system('cfn-lint', f.path).nil?
           puts "cfn-lint run complete"
         end
--- a/lib/stack_master/commands/lint.rb
+++ b/lib/stack_master/commands/lint.rb
@@ -1,0 +1,42 @@
+require 'tempfile'
+
+module StackMaster
+  module Commands
+    class Lint
+      include Command
+      include Commander::UI
+
+      def initialize(config, stack_definition, options = {})
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def perform
+        unless cfn_lint_available
+          puts "Failed to run cfn-lint, do you have it installed and available in $PATH?"
+          return
+        end
+
+        Tempfile.open(['stack', '.json']) do |f|
+          f.write(proposed_stack.template_body)
+          system('cfn-lint', f.path).nil?
+          puts "cfn-lint run complete"
+        end
+      end
+
+      private
+
+      def stack_definition
+        @stack_definition ||= @config.find_stack(@region, @stack_name)
+      end
+
+      def proposed_stack
+        @proposed_stack ||= Stack.generate(stack_definition, @config)
+      end
+
+      def cfn_lint_available
+        !system('cfn-lint', '--version').nil?
+      end
+    end
+  end
+end
--- a/lib/stack_master/commands/lint.rb
+++ b/lib/stack_master/commands/lint.rb
@@ -20,7 +20,7 @@ module StackMaster
         Tempfile.open(['stack', ".#{proposed_stack.template_format}"]) do |f|
           f.write(proposed_stack.template_body)
           f.flush
-          system('cfn-lint', f.path).nil?
+          system('cfn-lint', f.path)
           puts "cfn-lint run complete"
         end
       end
--- a/lib/stack_master/commands/lint.rb
+++ b/lib/stack_master/commands/lint.rb
@@ -17,7 +17,7 @@ module StackMaster
           return
         end
 
-        Tempfile.open(['stack', '.json']) do |f|
+        Tempfile.open(['stack', ".#{proposed_stack.template_format}"]) do |f|
           f.write(proposed_stack.template_body)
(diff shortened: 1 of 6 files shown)


viraptor:
Failure:
```
Executing lint on market-splunk-indexer in us-east-1
Failed to run cfn-lint, do you have it installed and available in $PATH?
```

Success:
```
Executing lint on market-splunk-indexer in us-east-1
cfn-lint 0.3.3
E2002 Parameter MarketAppVpcNatGatewayIps has invalid type List<String>
/var/folders/kq/pm2zcyfs5sggbt9sc5jcmrnm0000gn/T/stack20180711-72036-heg8pd.json:9:7

...
```

lib/stack_master/commands/lint.rb
```diff
@@ -0,0 +1,42 @@
...
+        @stack_definition = stack_definition
+      end
+
+      def perform
+        unless cfn_lint_available
+          puts "Failed to run cfn-lint, do you have it installed and available in $PATH?"
+          return
+        end
+
+        Tempfile.open(['stack', '.json']) do |f|
```
patrobinson: What if it's a YAML template?

viraptor: I was under the impression that proposed_stack.template_body will always be json. Is that incorrect?

If it can be yaml, that's supported too, just need to change the suffix.

stevehodgkiss: Yeah the template sent to CF is always JSON from StackMaster (even if the original template is YAML).

lib/stack_master/commands/lint.rb
```diff
@@ -0,0 +1,42 @@
...
+          f.write(proposed_stack.template_body)
```
stevehodgkiss: An alternative to using files could be to send the template as STDIN to cfn-lint (assuming it supports that).

lib/stack_master/commands/lint.rb
patrobinson: AFAIK we support native YAML templates and we submit them as YAML to CloudFormation. I do not know though if template_body will always be json.

viraptor:
I'd be happy with the compile command instead. cfn-lint doesn't do stdin, but you can always use `/dev/stdin` instead.

I'd prefer the `lint` one instead for community reasons: It's not a well known project, but has a lot of potential. If it was explicitly raised as a possibility from the help screen, I think that would improve life for a lot of people. (if I didn't have to wait for AWS to break after submitting the stack, I'd save so much time...)

But I can go with consensus. Are the strong feelings about the approach either way?

stevehodgkiss:
I'm not strongly opposed to including a `lint` command. Happy for it to be included if it would be useful. I think a `compile` command would be a good addition regardless of this choice though.

simpsora:
:+1: to `compile`.
:+1: to linting.  Given that a good portion of SM's functionality is transforming templates from one format to the other (and hence you're often not working in the native format), it'd be great to have an offline way to ensure the generated template is valid.

lib/stack_master/commands/lint.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      def initialize(config, stack_definition, options = {})
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def perform
+        unless cfn_lint_available
+          puts "Failed to run cfn-lint, do you have it installed and available in $PATH?"
+          return
```
patrobinson: This should exit with a non-zero exit code.

lib/stack_master/commands/lint.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+
+        Tempfile.open(['stack', ".#{proposed_stack.template_format}"]) do |f|
+          f.write(proposed_stack.template_body)
+          f.flush
+          system('cfn-lint', f.path).nil?
+          puts "cfn-lint run complete"
```
patrobinson: Can we pass the exit code from cfn-lint back?

lib/stack_master/commands/lint.rb
```diff
@@ -0,0 +1,43 @@
...
+      def perform
+        unless cfn_lint_available
+          puts "Failed to run cfn-lint, do you have it installed and available in $PATH?"
+          return
+        end
+
+        Tempfile.open(['stack', ".#{proposed_stack.template_format}"]) do |f|
+          f.write(proposed_stack.template_body)
+          f.flush
+          system('cfn-lint', f.path).nil?
```
patrobinson: What's the point of the nil check here?

lib/stack_master/commands/lint.rb
viraptor: cf-lint doesn't care :( it's always 0, whether it fails or not.

lib/stack_master/commands/lint.rb
viraptor: copy-pasta

lib/stack_master/commands/lint.rb
patrobinson: Boo

PR #249, opened 2018-08-17
Ensure Global Options take effect

### LINKED ISSUE #248: Command options not passed through when args are present
When `commander` is bumped to `v4.4.6` commands do not respect global options when args are also passed in. It must be related to https://github.com/commander-rb/commander/pull/72, but I'm not sure why it only triggers when args are passed in. Possibly because of https://github.com/commander-rb/commander/pull/72/files#diff-f5bc2065042b5aadd9785fa959c5af93R163? I don't know much about `commander` (and maybe this issue should go to them). For now I will pin `commander` to `v4.4.5`, but if someone can point me in the right direction I will gladly submit a patch.

Note that this applies to all commands.

Working example (no args, just global options):
```
$ AWS_PROFILE=dev bundle exec stack_master apply -c foo.yml -n
Executing apply on foo in us-west-2
Stack diff:
--- /var/folders/7r/82zfp0bs31b9pvnnydnwhsw80000gp/T/diffy20180814-12420-1jtv6b3        2018-08-14 15:54:33.000000000 -0500
+++ /var/folders/7r/82zfp0bs31b9pvnnydnwhsw80000gp/T/diffy20180814-12420-ykgu2j 2018-08-14 15:54:33.000000000 -0500
@@ -0,0 +1,7 @@
+{
+  "Resources": {
+    "IamUser": {
+      "Type": "AWS::IAM::User"
+    }
+  }
+}
\ No newline at end of file
Parameters diff:
+--- {}
No stack found

========================================
Proposed change set:
Add AWS::IAM::User IamUser
========================================
Create stack (y/n)? n
Stack creation aborted
```
Not working example (when args are present):
```
$ AWS_PROFILE=dev bundle exec stack_master apply us-west-2 foo -c foo.yml -n
Failed to load config file stack_master.yml
```

patrobinson:
Downgrade Commander to fix apparent bug in the parsing of Global Options when arguments are passed in.

Fixes #248

Without this PR the following feature test fails:

```
    And the output should not contain all of these lines:                 # aruba-0.14.6/lib/aruba/cucumber/command.rb:356
      | Stack diff:                 |
      | -    "TestSg2": {           |
      | Parameters diff: No changes |
      expected [#<Aruba::Command(#<Aruba::Processes::InProcess:70193980971960 commandline="stack_master --changed apply us-east-1 myapp-vpc --trace": output="Executing apply on myapp-vpc in us-east-1
      Stack diff: No changes
      Parameters ...">)>] not to include an object have output: string includes: "Stack diff:" (RSpec::Expectations::ExpectationNotMetError)
      features/apply.feature:303:in `And the output should not contain all of these lines:'
    Then the exit status should be 0                                      # aruba-0.14.6/lib/aruba/cucumber/command.rb:277
```

When I invested this a while ago I came to the (incorrect) conclusion that the test was badly written and not indicative of actual behaviour. I think we need a better test to ensure global options are respected. I'm not entirely sure what this test is actually doing, since the README says changed should be used without a stack argument (i.e. `stack_master --changed apply`)

We also should raise this upstream to the Commander repository.

### DIFF
--- a/stack_master.gemspec
+++ b/stack_master.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "timecop"
   spec.add_dependency "os"
   spec.add_dependency "ruby-progressbar"
-  spec.add_dependency "commander"
+  spec.add_dependency "commander", "<= 4.4.5"
   spec.add_dependency "aws-sdk-acm", "~> 1"
   spec.add_dependency "aws-sdk-cloudformation", "~> 1"
   spec.add_dependency "aws-sdk-ec2", "~> 1"


stack_master.gemspec
```diff
@@ -39,7 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "timecop"
   spec.add_dependency "os"
   spec.add_dependency "ruby-progressbar"
-  spec.add_dependency "commander"
+  spec.add_dependency "commander", "<= 4.4.5"
```
stevehodgkiss: Maybe `~> 4.4` would be better than anything less than this specific version?

patrobinson: Sorry I didn't specify in this PR, the issue was introduced in `4.4.6`

patrobinson: Ah yeah we can do something like `  spec.add_dependency "commander", "~> 4.4", "<= 4.4.5"`

Issue #237, opened 2018-05-31
Ability to set a timeout for cloudformation updates


markus7811:
If you use Cloudformation to deploy for example ECS Services, it works really nice! But if you have a failure within your docker image or your configuration and your container won't become healthy, cloudformation will try to get up the new container till it runs in its own timeout. 

But this timeout is about 3 hrs in my case. About 100 containers started and failed in this case. 

It would be nice to add a timeout per stack you can configure and if the cloudformation update process takes longer than a customizable timeout, it should call the aws api and cancel the update and exit with an code > 0 with message that timeout exceeded.

patrobinson:
My personal opinion is CloudFormation is the wrong choice for deploying containers, and any application, for this precise reason. It does not have the capabilities to define a state machine such "if this thing fails, rollback, or do some other thing". It's very limited (rollback or don't). Cancelling a stack update doesn't always work, you can end up in a state where the old container won't start and you're stuck waiting for the hardcoded 1 hour timeout. When it eventually fails you must delete the whole stack.

Internally we use ruby scripts to orchestrate container deployments.

patrobinson:
As previously stated, I think there are better tools than Cloudformation to deploy containers with and I don't think we should attempt to work around those limitations in StackMaster.

PR #250, opened 2018-08-22
Create latest_container resolver

patrobinson:
While we generally don't recommend deploying containers with CloudFormation, see #237, it can sometimes be useful.

This PR creates a parameter resolver that finds the latest image in an ECR repository and returns the full image path to that image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,8 +333,8 @@ Any value can be an array of possible matches.
 
 ### Latest Container from Repository
 
-Looks up the latest Container Image from an ECR repository. We use this instead of a "latest" tag, because if the tag is latest a CloudFormation update will not trigger a deployment of the container, since nothing has changed.
-As such this resolver will never return the latest tag for a container.
+Looks up the latest Container Image from an ECR repository. We use this instead of a "latest" tag, because if the tag is "latest" a CloudFormation update will not trigger a deployment of the container, since nothing has changed.
+As such this resolver will never return the tag "latest". If no other tag exists on the most recent container, it will return nil.
 
 ```yaml
 container_image_id:
--- a/README.md
+++ b/README.md
@@ -333,10 +333,12 @@ Any value can be an array of possible matches.
 
 ### Latest Container from Repository
 
-Looks up the latest Container Image from an ECR repository. We use this instead of a "latest" tag, because if the tag is "latest" a CloudFormation update will not trigger a deployment of the container, since nothing has changed.
-As such this resolver will never return the tag "latest". If no other tag exists on the most recent container, it will return nil.
+Looks up the a Container Image from an ECR repository. By default this will return the latest container in a repository.
+If `tag` is specified we return the sha digest of the image with this tag.
+This avoids the issue where CloudFormation won't update a Task Definition if we use a tag such as `latest`, because it only updates resources if a parameter has changed.
+This allows us to tag an image and deploy the latest version of that image via CloudFormation and avoids versioning our tags and having to store the metadata about the latest tag version somewhere.
 
-Returns the docker repository URI, i.e. `aws_account_id.dkr.ecr.region.amazonaws.com/container:tag`
+Returns the docker repository URI, i.e. `aws_account_id.dkr.ecr.region.amazonaws.com/container@sha256:digest`
 
 ```yaml
 container_image_id:
@@ -344,6 +346,7 @@ container_image_id:
     repository_name: nginx # Required. The name of the repository
     registry_id: 012345678910 # The AWS Account ID the repository is located in. Defaults to the current account's default registry
     region: us-east-1 # Defaults to the region the stack is located in
+    tag: production # By default we'll find the latest image pushed to the repository. If tag is specified we return the sha digest of the image with this tag
 ```
 
 ### Environment Variable
--- a/README.md
+++ b/README.md
@@ -331,6 +331,19 @@ A set of possible attributes is available in the [AWS documentation](https://doc
 
 Any value can be an array of possible matches.
 
+### Latest Container from Repository
+
+Looks up the latest Container Image from an ECR repository. We use this instead of a "latest" tag, because if the tag is latest a CloudFormation update will not trigger a deployment of the container, since nothing has changed.
+As such this resolver will never return the latest tag for a container.
+
+```yaml
+container_image_id:
+  latest_container:
+    repository_name: nginx # Required. The name of the repository
+    registry_id: 012345678910 # The AWS Account ID the repository is located in. Defaults to the current account's default registry
+    region: us-east-1 # Defaults to the region the stack is located in
+```
+
 ### Environment Variable
 
 Lookup an environment variable:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -331,6 +331,19 @@ A set of possible attributes is available in the [AWS documentation](https://doc
 
 Any value can be an array of possible matches.
 
+### Latest Container from Repository
+
+Looks up the latest Container Image from an ECR repository. We use this instead of a "latest" tag, because if the tag is latest a CloudFormation update will not trigger a deployment of the container, since nothing has changed.
+As such this resolver will never return the latest tag for a container.
```
mipearson: That sounds like a fun thing to find out the hard way.

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,49 @@
...
+        if parameters['repository_name'].nil?
+          raise ArgumentError, "repository_name parameter is required but was not supplied"
+        end
+        images = fetch_images(parameters['repository_name'], parameters['registry_id'], ecr_client)
+        return nil if images.empty?
+        images.sort! { |image_x, image_y| image_y.image_pushed_at <=> image_x.image_pushed_at }
+        latest_image = images.first
+        latest_tag = latest_image.image_tags.delete_if { |tag| tag == "latest" }.first
+        # aws_account_id.dkr.ecr.region.amazonaws.com
+        return  "#{latest_image.registry_id}.dkr.ecr.#{@region}.amazonaws.com/#{parameters['repository_name']}:#{latest_tag}"
```
viraptor: Is the tag safe to use here? Wouldn't we be better with `image_digest`?

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,49 @@
...
+            next_token: next_token,
+          })
+
+          images.push(resp.image_details)
+          next_token = resp.next_token
+
+          if resp.next_token.nil?
+            break
+          end
+        end
```
stevehodgkiss: Style nitpick:
```
begin
  # ..
end while next_token.nil?
```

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,49 @@
...
+        images = []
+        next_token = nil
+        while
+          resp = ecr.describe_images({
+            repository_name: repository_name,
+            registry_id: registry_id,
+            next_token: next_token,
+          })
+
+          images.push(resp.image_details)
```
stevehodgkiss: It looks like this will create an array of arrays containing `image_details` responses https://docs.aws.amazon.com/sdkforruby/api/Aws/ECR/Client.html#describe_images-instance_method

Maybe `images += resp.image_details`?

lib/stack_master/parameter_resolvers/latest_container.rb
patrobinson: This breaks one of my tests... and not the pagination test I just added. It seems to break stub_responses from the sdk...

lib/stack_master/parameter_resolvers/latest_container.rb
patrobinson: Hmm, recent events have made me question this view.
It seems as though we use our repositories not just to store the latest images but also as a cache for our CI pipelines. It's quite common for pipelines to push up images from a test environment. So finding the latest image is probably a terrible idea and what we should be doing instead is tagging latest, but fetching the sha for latest and returning that.

lib/stack_master/parameter_resolvers/latest_container.rb
stevehodgkiss: Ah, looks like it should be `!next_token.nil?`

patrobinson:
@stevehodgkiss @mipearson @viraptor sorry I changed the interface and code a bit based on my renewed understanding of my use case. We now allow specifying a tag parameter and returning the sha of that tag, to avoid having to version tags and store the latest tag version somewhere.

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+        images = fetch_images(parameters['repository_name'], parameters['registry_id'], ecr_client)
+        return nil if images.empty?
+        if !parameters['tag'].nil?
+          images.select! { |image| image.image_tags.any? { |tag| tag == parameters['tag'] } }
+        end
+        images.sort! { |image_x, image_y| image_y.image_pushed_at <=> image_x.image_pushed_at }
+        latest_image = images.first
+        # aws_account_id.dkr.ecr.region.amazonaws.com/repository@sha256:digest
+        return  "#{latest_image.registry_id}.dkr.ecr.#{@region}.amazonaws.com/#{parameters['repository_name']}@#{latest_image.image_digest}"
```
mipearson: Unnecessary return

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,51 @@
...
+    class LatestContainer < Resolver
+      array_resolver class_name: 'LatestContainers'
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(parameters)
+        @region = parameters['region'] || @stack_definition.region
```
mipearson: Maybe space this function out a bit to make it easier to understand logical groupings, eg:

```ruby
      def resolve(parameters)
        if parameters['repository_name'].nil?
          raise ArgumentError, "repository_name parameter is required but was not supplied"
        end

        @region = parameters['region'] || @stack_definition.region
        ecr_client = Aws::ECR::Client.new(region: @region)

        images = fetch_images(parameters['repository_name'], parameters['registry_id'], ecr_client)
        return nil if images.empty?

        if !parameters['tag'].nil?
          images.select! { |image| image.image_tags.any? { |tag| tag == parameters['tag'] } }
        end
        images.sort! { |image_x, image_y| image_y.image_pushed_at <=> image_x.image_pushed_at }
        latest_image = images.first

        # aws_account_id.dkr.ecr.region.amazonaws.com/repository@sha256:digest
        "#{latest_image.registry_id}.dkr.ecr.#{@region}.amazonaws.com/#{parameters['repository_name']}@#{latest_image.image_digest}"
      end
```

lib/stack_master/parameter_resolvers/latest_container.rb
```diff
@@ -0,0 +1,55 @@
...
+    class LatestContainer < Resolver
+      array_resolver class_name: 'LatestContainers'
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(parameters)
+        if parameters['repository_name'].nil?
```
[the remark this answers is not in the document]
patrobinson: No, as per the documentation it's only needed if you're fetching a container from a different account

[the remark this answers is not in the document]
andrewjhumphrey: NM, I see the default now.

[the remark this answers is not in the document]
andrewjhumphrey: Actually, no I don't, the docs say it should default but I don't see the defaulting happening.
And no test for what happens when it's not...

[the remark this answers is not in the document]
patrobinson: The API handles this for us. If we don't specify a registry ID, it defaults to the current account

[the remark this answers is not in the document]
andrewjhumphrey: Perfect, no need for tests then either :-)

PR #251, opened 2018-08-31
Print backtrace when compilation fails

patrobinson:
When a template fails to compile it's common to get an error like:

```
Executing apply on samson-db in us-east-1
StackMaster::TemplateCompiler::TemplateCompilationFailed Failed to compile /Users/matt/Documents/GitHub/our-samson/aws/templates/samson_db.rb.
 Caused by: NameError uninitialized constant #<Class:0x007f850a560748>::Sparkleformation 
```

Which is quite unhelpful to understand where the error occurs. This PR makes two changes to combat this:

- When we do catch an exception, print the error and backtrace.

### DIFF
--- a/lib/stack_master/template_compiler.rb
+++ b/lib/stack_master/template_compiler.rb
@@ -6,8 +6,8 @@ module StackMaster
       compiler = template_compiler_for_file(template_file_path, config)
       compiler.require_dependencies
       compiler.compile(template_file_path, compile_time_parameters, compiler_options)
-    rescue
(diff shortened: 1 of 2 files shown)


stevehodgkiss:
> Only catch StandardError. NameError and SyntaxError will throw an exception and not get caught

`StandardError` is the default for `rescue` so this isn't changing which errors are being caught.

patrobinson:
@stevehodgkiss you're right. StandardError will catch NameError but not SyntaxError. I'm not sure what we want to do so I'm just going with printing a backtrace

PR #255, opened 2018-10-17
Deprecate dotpg

patrobinson:
This is both a proposal and pull request. I propose that for version 2 of StackMaster we should remove dotgpg support. Dotgpg is difficult to use for a couple of reasons, firstly anytime someone joins or leaves our team we have to re-encrypt our secrets. Secondly there's a [long running issue](https://github.com/ConradIrwin/dotgpg/issues/20) where dotgpg will fail if there's an expired key.

Also removing Dotgpg will simplify our release process, which at the moment requires that we build two seperate versions one of which has to be built on a Windows machine.

This PR prints a deprecation warning when the secrets parameter resolver is used and updates the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,11 @@ stack_defaults:
   role_arn: service_role_arn
 region_defaults:
   us-east-1:
-    secret_file: production.yml.gpg
     tags:
       environment: production
     notification_arns:
       - test_arn
   ap-southeast-2:
-    secret_file: staging.yml.gpg
     tags:
       environment: staging
 stacks:
@@ -211,7 +209,9 @@ into parameters of dependent stacks.
 
 ### Secret
 
-Note: This resolver is not supported on Windows, you can instead use the [Parameter Store](#parameter-store).
+**DEPRECATED**: This parameters resolver is deprecated and will be removed in version 2.0. Refer to [Parameter Store](#parameter-store) and [1Password](#1password-lookup) for alternatives for handling secrets.
+
+Note: This resolver is not supported on Windows.
 
 The secret parameters resolver expects a `secret_file` to be defined in the
 stack definition which is a GPG encrypted YAML file. Once decrypted and parsed,
@@ -256,7 +256,9 @@ you will likely want to set the parameter to NoEcho in your template.
 db_password:
   parameter_store: ssm_parameter_name
 ```
+
 ### 1Password Lookup
+
 An Alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
 You declare a 1password lookup with the following parameters in your parameters file:
 
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ into parameters of dependent stacks.
 
 ### Secret
 
-**DEPRECATED**: This parameters resolver is deprecated and will be removed in version 2.0. Refer to [Parameter Store](#parameter-store) and [1Password](#1password-lookup) for alternatives for handling secrets.
+**DEPRECATED**: This parameters resolver is deprecated and will be removed in version 2.0. You may install the [seperate gem](https://github.com/envato/stack_master-gpg_parameter_resolver) or use [Parameter Store](#parameter-store) or [1Password](#1password-lookup).
 
 Note: This resolver is not supported on Windows.
 
--- a/features/apply_with_secret_parameters.feature
+++ b/features/apply_with_secret_parameters.feature
@@ -1,0 +1,51 @@
+Feature: Apply command with parameter_store parameter
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us-east-2:
+          vpc:
+            template: vpc.rb
+            secret_file: production.yml.gpg
+      """
+    And a directory named "parameters"
+    And a file named "parameters/vpc.yml" with:
+      """
+      vpc_cidr:
+        secrets: "cucumber-test-vpc-cidr"
+      """
+    And a directory named "secrets"
+    And a file named "secrets/production.yml.gpg" with ""
+    And a directory named "templates"
+    And a file named "templates/vpc.rb" with:
+      """
+      SparkleFormation.new(:vpc) do
+
+        parameters.vpc_cidr do
+          type 'String'
+        end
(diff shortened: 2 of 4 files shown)


jacobbednarz:
> Dotgpg is difficult to use for a couple of reasons, firstly anytime someone joins or leaves our team we have to re-encrypt our secrets. 

This is quite a handy feature of GPG in general and I don't think a lack of on/off boarding process should warrant a removal of it. This ensures that even if someone manages to access sensitive data after they have left the team, they cannot decrypt.

> ... Secondly there's a [long running issue](https://github.com/ConradIrwin/dotgpg/issues/20) where dotgpg will fail if there's an expired key.

Has anyone submitted a PR for this? I see we raised it but no actions?

> Also removing Dotgpg will simplify our release process, which at the moment requires that we build two seperate versions one of which has to be built on a Windows machine.

Travis CI now supports [Windows CI](https://docs.travis-ci.com/user/reference/windows/) so this process could be automated based on tagging or master merge if you wanted. Regardless of this PR, I'd propose we move to automating the release process seeing how [Rubygems](https://docs.travis-ci.com/user/deployment/rubygems/) and [GitHub releases](https://docs.travis-ci.com/user/deployment/releases/) are both accessible out of the box. 

I'm not particularly sold on Parameter Store or 1Password as the only ongoing solution for secrets primarily for the fact they both have usage caveats that dotgpg and local encrypted storage addressed.

viraptor:
For a better discussion, could you expand this?

> both have usage caveats that dotgpg and local encrypted storage addressed.

Regarding access after someone leaves the team, don't PS and 1Password address this with permissions already? In the first case, AWS account/permissions are dropped. In the second, team membership is revoked.
I think the access revocation is actually better in this situation - if anyone gets access to the encrypted version, they likely get access to the whole `git` repo, not just the last commit (either via someone's workstation or remote repo access). That means they can forever access the old commits using their old key - we'd need to roll credentials every time people leave to prevent this.

jacobbednarz:
@viraptor The usage caveats aren't anything to do with on/off boarding permissioning, they are related to the integration and service usage. For instance, 1Password CLI is only available for subscription based services and parameter store requires managing KMS + carefully avoiding the request/size limitations. Last POC we did using parameter store for replacing an existing system seen us hitting these limitations when spinning up reasonably sized fleets. Both of these are hurdles to anyone wanting to implement stack master with sensitive data either for themselves or a small team. 

> In the second, team membership is revoked.

Only if it can phone home 😉 If the machine isn't connected to the internet, this isn't something that happens.

> That means they can forever access the old commits using their old key - we'd need to roll credentials every time people leave to prevent this.

Jaz looks after rolling most of our shared credentials and teams should be looking after their own credential management which includes a regular rotation of credentials so this should be mitigated.

patrobinson:
>  parameter store requires managing KMS + carefully avoiding the request/size limitations. Last POC we did using parameter store for replacing an existing system seen us hitting these limitations when spinning up reasonably sized fleets.

StackMaster isn't likely to hit the same limitations because you only access KMS and ParameterStore when you create/update the stack, not every time you spin up a new ec2 instance.

We do not roll keys today when somebody leaves, we just re-encrypt them. So I don't see how removing GPG access creates more risk. It's the same risk with the benefit of not having to manage the encryption keys ourselves, which is a net positive to me.

simpsora:
👍 to deprecating the `dotgpg` resolver, but I'd like it if there were some other offline mechanism.  Both the 1P and PS resolvers require access to external systems, which isn't ideal.

Is an `ejson-`-based resolver a possibility?  Would it have the same issues that the `dotgpg` one does?

patrobinson:
@simpsora No, I think with https://github.com/envato/ejson_wrapper we could make that work especially since there are already EJSON ruby bindings.

stevehodgkiss:
Yeah an EJSON secret resolver would be easy enough to write using EJSON wrapper. Access to decrypt secrets could be managed via permissions on the KMS key EJSON wrapper uses (to encrypt EJSON's actual private key).

patrobinson:
Awesome. Anybody volunteering to write the parameter resolver 😃 ?

orien:
Is it possible to extract the Dotgpg secret resolver into a separate (and optional) gem for those who may want to continue using it?

patrobinson:
@orien that would be possible and is already supported by StackMaster

orien:
I've created [envato/stack_master-gpg_parameter_resolver](https://github.com/envato/stack_master-gpg_parameter_resolver) and extracted the secret parameter resolver. Some lightweight tests prove this works fine.

Is the name ok?

viraptor:
Just an FYI: there's a new tool being built for file encryption based on SSH keys with nice integration to github keys and much of the complexity of gpg removed. Worth a read: https://docs.google.com/document/d/11yHom20CrsuX8KQJXBBw04s80Unjv8zCg_A7sPAX_9Y/view#
Could be a good replacement once released.

patrobinson:
@orien looks great!

stevehodgkiss:
Nice one @orien! The name seems ok. I think ideally we would have named this resolver `gpg` to begin with and used `gpg: my_secret` rather than `secret: my_secret`.

Can we remove the secret resolver from StackMaster with a fallback message to install https://github.com/envato/stack_master-gpg_parameter_resolver ?

patrobinson:
I think that'd be a breaking change and have to wait till 2.0. But we can certainly add here a link to the gem

patrobinson:
Given there's an alternative offline mechanism (ejson) and GPG is now available via a seperate Gem I am going to merge this unless there are objections.

README.md
```diff
@@ -211,7 +209,9 @@ into parameters of dependent stacks.
 
 ### Secret
 
-Note: This resolver is not supported on Windows, you can instead use the [Parameter Store](#parameter-store).
+**DEPRECATED**: This parameters resolver is deprecated and will be removed in version 2.0. You may install the [seperate gem](https://github.com/envato/stack_master-gpg_parameter_resolver) or use [Parameter Store](#parameter-store) or [1Password](#1password-lookup).
```
orien: Let's provide the option for EJSON+KMS also.

PR #264, opened 2018-12-21
Add a parameter resolver for EJSON wrapper

stevehodgkiss:
An EJSON file that uses KMS can be generated with EJSON Wrapper:

```
ejson_wrapper generate --region ap-southeast-2 --kms-key-id [key_id] --file secrets/staging.ejson
```

Secrets are then added as usual with EJSON. Add the plaintext then:

```
ejson encrypt secrets.ejson
```

Set `ejson_file` in the stack definition to `staging.ejson`. Then secrets can be referenced in parameter files:

```
my_param:
  ejson: "my_secret"
```

TODO:

- [x] Update Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,11 @@ my_param:
   ejson: "my_secret"
 ```
 
+Additional configuration options:
+
+- `ejson_file_region` The AWS region to attempt to decrypt private key with
+- `ejson_file_kms` Default: true. Set to false to use ejson without KMS.
+
 ### Security Group
 
 Looks up a security group by name and returns the ARN.
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ key_name: myapp-us-east-1
 
 ### Compile Time Parameters
 
-Compile time parameters can be used for [SparkleFormation](http://www.sparkleformation.io) templates. It conforms and 
+Compile time parameters can be used for [SparkleFormation](http://www.sparkleformation.io) templates. It conforms and
 allows you to use the [Compile Time Parameters](http://www.sparkleformation.io/docs/sparkle_formation/compile-time-parameters.html) feature.
 
 A simple example looks like this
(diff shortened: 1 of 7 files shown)


lib/stack_master/parameter_resolvers/ejson.rb
```diff
@@ -0,0 +1,46 @@
...
+        @stack_definition.ejson_file
+      end
+
+      def ejson_file_path
+        @ejson_file_path ||= File.join(@config.base_dir, secret_path_relative_to_base)
+      end
+
+      def secret_path_relative_to_base
+        @secret_path_relative_to_base ||= File.join('secrets', ejson_file)
+      end
```
patrobinson: These two methods have no effect and are not called anywhere? When I specify `ejson_file` I have to use the full path (`secrets/foo.ejson`)

README.md
```diff
@@ -274,6 +275,29 @@ Currently we support two types of secrets, `password`s and `secureNote`s. All va
 
 For more information on 1password cli please see [here](https://support.1password.com/command-line-getting-started/)
 
+### EJSON Store
+
+[ejson](https://github.com/Shopify/ejson) is a tool for managing asymmetrically encrypted values in JSON format. This allows you to keep secrets securely in git/Github and give anyone the ability the capability to add new secrets without requiring access to the private key.
+
+First create the `production.ejson` file and store the secret value in it, then call `ejson encrypt secrets.ejson`. Then add the `ejson_file` argument to your stack in stack_master.yml:
```
stevehodgkiss: Shall we provide details about how to create this file? For example: 

```
ejson_wrapper generate --region ap-southeast-2 --kms-key-id [key_id] --file secrets/production.ejson
```

lib/stack_master/parameter_resolvers/ejson.rb
stevehodgkiss: I've updated the resolver to look in `secrets/` for the `ejson_file`

README.md
stevehodgkiss: We could add another command to stack master to create ejson files with ejson wrapper/a kms key rather than relying on the user to know about ejson wrapper and do this manually. Thoughts?

patrobinson: Yeah I think that's a nice addition

lib/stack_master/parameter_resolvers/ejson.rb
```diff
@@ -0,0 +1,46 @@
...
+      SecretNotFound = Class.new(StandardError)
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+      end
+
+      def resolve(secret_key)
+        validate_ejson_file_specified
+        secrets = decrypt_ejson_file
```
joneslee85: @stevehodgkiss should we memoize secrets?

PR #271, opened 2019-01-03
Test on latest version of Rubygems and Bundler

orien:
And loosen constraints to allow Stack Master developers to use the newly released Bundler 2.0! https://bundler.io/blog/2019/01/03/announcing-bundler-2.html#

### DIFF
--- a/stack_master.gemspec
+++ b/stack_master.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.1.0"
   spec.platform      = gem_platform
 
-  spec.add_development_dependency "bundler", "~> 1.5"
+  spec.add_development_dependency "bundler"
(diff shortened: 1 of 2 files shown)


stack_master.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.1.0"
   spec.platform      = gem_platform
 
-  spec.add_development_dependency "bundler", "~> 1.5"
+  spec.add_development_dependency "bundler"
```
patrobinson: Should we just change this to `> 1.5` to avoid having to support earlier versions?

orien: It's just a development dependency, this change won't influence the released gem.

I think there's only a minute chance that there're Stack Master developers with such a old version of Bundler installed.

PR #272, opened 2019-01-10
Add quietly flag

patrobinson:
Surpresses stack events and just returns to the CLI

### DIFF
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -94,6 +94,10 @@ module StackMaster
     autoload :Streamer, 'stack_master/stack_events/streamer'
   end
 
+  NON_INTERACTIVE_DEFAULT = false
+  DEBUG_DEFAULT = false
+  QUIET_DEFAULT = false
+
   def interactive?
     !non_interactive?
   end
@@ -101,7 +105,7 @@ module StackMaster
   def non_interactive?
     @non_interactive
   end
-  @non_interactive = false
+  @non_interactive = NON_INTERACTIVE_DEFAULT
 
   def non_interactive!
     @non_interactive = true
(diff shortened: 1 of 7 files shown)


lib/stack_master.rb
```diff
@@ -121,6 +125,19 @@ def debug(message)
...
+    @quiet = true
+  end
+  @quiet = QUIET_DEFAULT
+
+  def quiet?
+    @quiet
+  end
+
+  def reset_flags
+    @quiet = QUIET_DEFAULT
```
patrobinson: Adding the following to this method:
```
    @non_interactive = NON_INTERACTIVE_DEFAULT
    @debug = DEBUG_DEFAULT
```

Results in random feature test failures...

```
    Then the stack "myapp-vpc" should contain this notification ARN "test_arn_1" # features/step_definitions/stack_steps.rb:54
      Stack does not exist (Aws::CloudFormation::Errors::ValidationError)
      ./lib/stack_master/test_driver/cloud_formation.rb:127:in `describe_stacks'
      ./features/step_definitions/stack_steps.rb:55:in `/^the stack "([^"]*)" should contain this notification ARN "([^"]*)"$/'
      features/stack_defaults.feature:77:in `Then the stack "myapp-vpc" should contain this notification ARN "test_arn_1"'
```

mipearson: Yeah, because it's a module variable it's persistent between tests - rspec doesn't unload/reload modules between individual tests, or even between files (god, could you imagine how slow that would be?!)

mipearson: workaround: add a `before(:each)` block to your `spec_helper` that resets these to defaults

patrobinson: From @stevehodgkiss 
> `non_interactive` is setup once when the test suite loads (on requiring `stack_master/testing`), so resetting it would undo that change.

Issue #235, opened 2018-05-31
Wrong exit code if apply fails because of wrong parameters


markus7811:
Hi,

i use stack_master in our gitlab-ci environment a lot. One thing I think is not good is the exit code of the apply command, if there are empty/blanc parameters.

```
$ bundle exec stack_master apply tst sfn-seec--tst--default--base -y
Executing apply on sfn-seec--tst--default--base in eu-central-1
Stack diff: No changes
**Empty/blank parameters detected, ensure values exist for those parameters. Parameters will be read from the following locations:**
 - /builds/Free4All/SparkleformationPacks/sfn_seec_base/generic/parameters/tst/sfn_seec__tst__default__base.yml
Parameters diff: 
+---
.... my parameters


**No stack found**
$ bundle exec stack_master delete tst sfn-seec--tst--default--base -y
Stack does not exist
Job succeeded
```

So i think this should fail.

patrobinson:
Sorry what is the exit code and what do you expect it to be?

markus7811:
For example here is one CF Parameter missing:

```
$ bundle exec stack_master apply "$ENVIRONMENT-$UNIQUE_IDENTIFIER" "$APPLICATION_NAME--$ENVIRONMENT--$UNIQUE_IDENTIFIER--service-$SERVICE" --config "stack_master_$ENVIRONMENT-$UNIQUE_IDENTIFIER.yml" -y --trace
Executing apply on blablabla-1
Stack diff: No changes
Empty/blank parameters detected, ensure values exist for those parameters. Parameters will be read from the following locations:
 - /home/markus/ ....
Parameters diff:
...

Build succeeded
markus@Markus-VirtualBox:~...$ echo $?
0
```

The exit code is 0 but there was an error - an empty or blanc parameter. So the stack_master apply command should return 1 instead

markus7811:
The same for:

```
bundle exec stack_master delete tst sfn-seec--tst--default--base -y
Stack does not exist
Job succeeded
```
if the stack was not found, it should return 1 instead of 0 because the execution was not possible

markus7811:
Ok, changing the behavior of apply was easy - but changing delete is tricky, because you call this method not the way you do for other commands

PR #273, opened 2019-01-31
Add option for setting sleep_between_fetches and cleaned up return codes for failed apply / delete

markus7811:
We are using stack_master a in our ci environment. This pull request would fix 2 of our major problems with stack_master.

1. stack_master makes a lot of api requests. If you are running a lot of stack_updates you can reach the api limit for the cloudformation api really easy. So i made it possible to set the `sleep_between_fetches` interval vial the stack_master.yml file (and changed the default value to 30 seconds - i think this would be really enough, as cloudformation takes some time to create resources and within pipelines you don't care for about 30 extra seconds for finishing the pipeline). This would lower the api calls by factor 30 and has even the ability to change this value.

2. I changed the exit codes for apply and delete operations, that fail. At the moment stack_master mostly exits with an exit code of 0 (everything ok). So our pipeline is allways green, even if the operations (apply, delete, ...) fail. Now, the success state of the commands is evaluated and stack_master exits with 1, if the success state is not true. 
Solves: https://github.com/envato/stack_master/issues/235

It would be nice, if you would accept these Request. Please tell me, what you want to be changed.
(and excuse my bad ruby and chaos commits - it's my first time working with github and ruby)


markus7811:
I was able to fix the rspec for apply, but have no idea at the moment how to fix the delete spec

markus7811:
ok, the tests that are now failing have nothing to do with my changes in my opinion :-)

markus7811:
ok, it was my mistake

markus7811:
So because now validating a broken template fails with 1 i think the cucumber test breaks - but i have no idea how to fix this behaviour

patrobinson:
Hey @markus7811 thanks for the PR!

Firstly these two problems (exit codes and api rate limiting) are seperate concerns and as such I'd like to see seperate PRs for them. This enables us to have a discussion about each one individually and not let any concerns about one feature prevent the other feature from being merged. I'm very happy to see the exit code be more descriptive and would be happy to help with failing tests to get that over the line.

Secondly, I don't think the approach of letting the user determine how long we sleep for is the best user experience. Instead I imagine solving this problem by catching rate limiting errors from the API and exponentially backing off and retrying, as [described in the AWS documentation](https://docs.aws.amazon.com/general/latest/gr/api-retries.html). This would allow us to deal with this without forcing the user to think about or tune a sleep interval.

PR #274, opened 2019-02-01
Feature/bugfix return codes

markus7811:
With this pull request i want to change the exit codes of stack_master for apply and delete, because they are 0 at the moment(everything ok) even if apply can't find parameters and delete can't find the stack to delete.

Now the exit code of this commands will return 1 if they fail.

We need this for using stack_master in cicd pipelines. At the moment the pipeline state is ok/green, even if some commands failed.

This belongs to issue https://github.com/envato/stack_master/issues/235


patrobinson:
Does this PR solve the issue? https://github.com/envato/stack_master/pull/276

patrobinson:
I think this is fixed in #280

PR #276, opened 2019-02-14
return non-zero exit status when command fails

grosser:
- the return value of execute_stacks_command is not getting used for anything by commander see https://github.com/commander-rb/commander/issues/78
 - a missing stack did not cause an error

If it were not for https://github.com/cucumber/aruba/issues/592 it could be simplified to `@kernel.exit command_results.all?`

### DIFF
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -144,6 +144,11 @@ Feature: Apply command
     And the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} myapp-web AWS::CloudFormation::Stack CREATE_COMPLETE/
     Then the exit status should be 0
 
+  Scenario: Run apply with invalid stack
+    When I run `stack_master apply foo bar`
+    Then the output should contain "Could not find stack definition bar in region foo"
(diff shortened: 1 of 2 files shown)


features/apply.feature
```diff
@@ -144,6 +144,11 @@ Feature: Apply command
     And the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} myapp-web AWS::CloudFormation::Stack CREATE_COMPLETE/
     Then the exit status should be 0
 
+  Scenario: Run apply with invalid stack
+    When I run `stack_master apply foo bar`
+    Then the output should contain "Could not find stack definition bar in region foo"
+    Then the exit status should be 1
```
patrobinson: very very minor niggle. I think the "Then" on this line should be "And"

grosser:
updated with the `And` ... who will merge this ?

patrobinson:
Released as 1.13.0

grosser:
please use `rake release` or something similar so tags get added, I tried to find v1.13.0 but could not

Issue #277, opened 2019-02-20
stack_master apply with no changes exits with non-zero


sergei-matheson:
In version 1.13, `stack_master apply` exits with status code 1 if there are no changes. It also exits with status code 1 from `stack_master diff` if there are no changes. These are breaking changes in behaviour from previous versions.

In my opinion, while returning non-zero from a diff command with no changes makes sense, returning non-zero from an apply command with no changes does not. This would mean, in a CI/CD environment, that a build that deployed infrastructure would pass, and then a rebuild, with no changes, would fail. Obviously, there would be ways to work around this, but it seems counter-intuitive.

patrobinson:
Related to https://github.com/envato/stack_master/pull/276

grosser:
... or add that `puts caller` to `def failed` since that seems to be how errors are added

PR #279, opened 2019-03-19
Return true if nothing has failed in cli

logingood:
We should ensure that we return true,
otherwise changes introduced in #276 will cause stack_master to always
exit with code 1, which breaks integration tests.

Addresses issues opened in #277.

cc: @grosser

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -234,6 +234,8 @@ module StackMaster
 
       # fail command execution if something went wrong
       @kernel.exit 1 unless command_results.all?
+
+      true
     end
   end
 end


grosser:
this looks wrong to me, the result of execute_stacks_command is not used anywhere afaik

grosser:
https://github.com/envato/stack_master/pull/276 made the execute_stacks_command do an `exit 1` when any `command.perform` failed ... so maybe the issue is that some `command.perform` do not return true ?
if `commander` fails when true is not returned, then the `exit 1` added in #276 should be removed

grosser:
yep ... lib/stack_master/commands/apply.rb does never return true

logingood:
so just remove OS exit from #276 @grosser and maybe supply a message that stack is missing instead ?

grosser:
The exit is correct in that position afaik, we want `exit 1` if a command fails and should not ignore the issue. For some reason an empty apply thinks it's failed ... most likely coming from lib/stack_master/command.rb:26 ... so add a `puts caller` there to see who triggers it ...

logingood:
it doesn't go in command.rb:26, will take further look

logingood:
@grosser it comes from here https://github.com/envato/stack_master/blob/master/bin/stack_master#L14, thing does `exit false`, which returns status code 1 if there is no `true` returned by `execute!` from cli.rb

logingood:
```⌂155% [mmukhtarov:~] 1 % ruby /tmp/test.rb; echo $?
1
⌂145% [mmukhtarov:~] % cat /tmp/test.rb
#!/usr/bin/ruby

def foo
  puts
end

result = foo
exit !!result
```

logingood:
well, I probably getting the problem wrong, anyway would appreciate any help, also will look into tests.

grosser:
https://github.com/envato/stack_master/pull/280

Issue #278, opened 2019-02-25
stack_master validate exits with non-zero on valid stack


orien:
`stack_master validate` exits with a non-zero code on a valid stack.

```
> stack_master --version
StackMaster 1.13.0
echo $?
1

> stack_master validate us-east-1 bastion-vpc
Executing validate on bastion-vpc in us-east-1
bastion-vpc: valid
> echo $?
1
```

This causes a bit of havoc in continuous integration where we rely on the exit code to indicate the validity of the stack.

orien:
This was potentially introduced in #276 and seems closely related to #277.

logingood:
https://github.com/envato/stack_master/pull/279

PR #280, opened 2019-03-19
no need for explicit exit since bin/stack_master already handles that

grosser:
followup to https://github.com/envato/stack_master/pull/276 which fixed 1 bug and introduced another 😨 
the cucumber test was misleading since it never executes bin/stack_master and so never gets the actual exit status
also simplifying by converting the array logic to a boolean

replaces https://github.com/envato/stack_master/pull/279
fixes https://github.com/envato/stack_master/issues/277
fixes https://github.com/envato/stack_master/issues/278

### DIFF
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -147,7 +147,7 @@ Feature: Apply command
   Scenario: Run apply with invalid stack
     When I run `stack_master apply foo bar`
     Then the output should contain "Could not find stack definition bar in region foo"
-    And the exit status should be 1
+    And the exit status should be 0
 
   Scenario: Create stack with --changed
     Given I stub the following stack events:
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -147,7 +147,6 @@ Feature: Apply command
   Scenario: Run apply with invalid stack
     When I run `stack_master apply foo bar`
     Then the output should contain "Could not find stack definition bar in region foo"
-    And the exit status should be 0
 
   Scenario: Create stack with --changed
     Given I stub the following stack events:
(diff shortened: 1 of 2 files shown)


features/apply.feature
```diff
@@ -147,7 +147,7 @@ Feature: Apply command
   Scenario: Run apply with invalid stack
     When I run `stack_master apply foo bar`
     Then the output should contain "Could not find stack definition bar in region foo"
-    And the exit status should be 1
+    And the exit status should be 0
```
patrobinson: Wouldn't this be 1 based on `if stack_definitions.empty? ... success = false` ?

logingood: it would be 1 if success returns false, because `exit !!false` in bin/stack_master.rb

grosser: it will never be 1 since we do not run `bin/stack_master` in this test, which includes the `exit`

grosser:
tests for `validate` can be their own PR, would prefer to get this out to get the bug fixed

features/apply.feature
patrobinson: I think we should just remove this test rather than introduce something that tests for undesirable behaviour.

grosser: then it should be removed from all of them for consistency 🤷‍♂️

patrobinson: While the other tests are not accurate they do reflect the behaviour we want to see.

This test is both inaccurate and does not reflect true behaviour

patrobinson:
```suggestion
```

grosser:
updated

patrobinson:
Released as 1.13.1

PR #282, opened 2019-06-16
Latest s3objectversion

crashdummymch:
This will pull the latest version of an s3 object which has version control.  This is utilized during lambda function/layer updates.  While this code does work.  It looks like stack_master performs template diff/expansion before the upload.  This causes the code to have the previous version_id instead of the one that stack_master uploads.  I could use some pointers on how to fix this please.


crashdummymch:
Found where the files upload is done in and moved that function to the first action performed in the apply command

lib/stack_master/parameter_resolvers/latest_s3_object_version.rb
```diff
@@ -0,0 +1,22 @@
...
+        @stack_definition = stack_definition
+        @s3_object_finder = S3ObjectFinder.new(@stack_definition.region)
+      end
+
+      def resolve(parameters)
+        parameters_map = @s3_object_finder.build_filters_from_hash(parameters)
+        s3object = @s3_object_finder.find_latest_file(parameters_map)
+        # debugging of s3object
+        # STDERR.puts "s3object type=#{s3object.class}"
+        # STDERR.puts "s3object[:version_id]=#{s3object.try(:version_id)}"
```
patrobinson: Can you please remove these debug statements?

lib/stack_master/parameter_resolvers/s3_object_finder.rb
```diff
@@ -0,0 +1,31 @@
...
+        # create hash with key converted to symbol
+        hash.map { |key, value| [key.to_sym, value.to_s] }.to_h
+      end
+
+      def find_latest_file(parameters)
+        # STDERR.puts parameters
+        file_versions = s3.list_object_versions(parameters)
+        # STDERR.puts file_versions.versions
+        sorted_file_versions = file_versions.versions.sort do |a, b|
+          # STDERR.puts "last_modified=#{a.last_modified.class}"
```
patrobinson: As above, debug statements should either be removed or use `StackMaster.debug` to only print them when the debug flag is set

lib/stack_master/parameter_resolvers/s3_object_finder.rb
```diff
@@ -0,0 +1,31 @@
+module StackMaster
+  module ParameterResolvers
+    class S3ObjectFinder
```
patrobinson: Can you help me understand why there are two classes? As far as I can tell this class is only ever used by the latest s3 object resolver, so I think it would be more idiomatic to keep this code together.

lib/stack_master/parameter_resolvers/latest_s3_object_version.rb
```diff
@@ -0,0 +1,22 @@
...
+      array_resolver class_name: 'LatestS3ObjectVersions'
+
+      def initialize(config, stack_definition)
+        @config = config
+        @stack_definition = stack_definition
+        @s3_object_finder = S3ObjectFinder.new(@stack_definition.region)
+      end
+
+      def resolve(parameters)
+        parameters_map = @s3_object_finder.build_filters_from_hash(parameters)
```
patrobinson: I would rather we solve the immediate use case and only take parameters we need, rather than solving for every use case by taking every possible parameter the S3 method can receive.
This would also help understand how this method can be used, right now it's not entirely clear what `parameters` needs to contain to get a result.
So can we limit the arguments `resolve()` takes to just those needed? (bucket and key?)

crashdummymch:
these are all good comments and will work towards it.  After further investigation, it still looks like it parses and instantiates the parameter files before the file uploads are done.  Therefore, I still need help figuring out how to ensure the stack parameters are re-parsed after the files are uploaded.  Any help would be appreciative.

lib/stack_master/parameter_resolvers/s3_object_finder.rb
crashdummymch: This would allow for multiple types of searching for s3 objects; similar to latest_ami_by_tags and latest_ami.  I plan to add additional resolvers once this is fully working.

patrobinson:
@crashdummymch I was thinking about this problem the other day and I think a better solution would be to implement this as something resolved by the CloudFormation template or DSL.

If you're using straight cloudformation this could be a SSM Parameter; if you're using SparkleFormation we would be interested in adding a custom ruby function that compiles and uploads Lambda code to S3 for you, then returns the s3 path to that file.

PR #283, opened 2019-06-25
Add ability to restrict which accounts can work with stacks

petervandoros:
In order to provide extra safety and reduce the risk of applying non-production changes to production, this PR adds the ability to specify which accounts are allowed to work with specific stacks.

The behaviour can be summarised using the following example stack_master.yml file:

```yaml
stack_defaults:
  allowed_accounts: '555555555'
stacks:
  us-east-1:
    myapp-vpc: # inherits allowed account 555555555
      template: myapp_vpc.rb
      tags:
        purpose: front-end
    myapp-db:
      template: myapp_db.rb
      allowed_accounts: # only these accounts
        - '1234567890'
        - '9876543210'
      tags:
        purpose: back-end
    myapp-web:
      template: myapp_web.rb
      allowed_accounts: [] # allow all accounts by overriding stack defaults
      tags:
        purpose: front-end
    myapp-redis:
      template: myapp_redis.rb
      allowed_accounts: '888888888' # only this account
      tags:
        purpose: back-end
```

#### Changes

- Add `allowed_accounts` stack definition property that lists the aws account IDs allowed to work with the stack
- Add a `--skip-account-check` flag to bypass the account checks
- Updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,7 +551,7 @@ Note though that if a dynamic with the same name exists in your `templates/dynam
 
 ## Allowed accounts
 
-The AWS account the command is executing in can be restricted to a specific list of allowed account. This is useful in reduicing the possibility of applying non-production changes in a production account. Each stack definition can specify the `allowed_accounts` property with an array of AWS account IDs the stack is allowed to work with.
+The AWS account the command is executing in can be restricted to a specific list of allowed accounts. This is useful in reducing the possibility of applying non-production changes in a production account. Each stack definition can specify the `allowed_accounts` property with an array of AWS account IDs the stack is allowed to work with.
 
 This is an opt-in feature which is enabled by specifying at least one account to allow.
 
--- a/README.md
+++ b/README.md
@@ -555,32 +555,32 @@ The AWS account the command is executing in can be restricted to a specific list
 
 This is an opt-in feature which is enabled by specifying at least one account to allow.
 
-Unlike other stack defaults, `allowed_accounts` values specified in the stack definition override values in the stack defaults instead of merging them together. This allows specifying allowed accounts in the stack defaults for all stacks and still have different values for specific stacks. See below example config for an example.
+Unlike other stack defaults, the `allowed_accounts` property values specified in the stack definition override values specified in the stack defaults (i.e., other stack property values are merged together with those specified in the stack defaults). This allows specifying allowed accounts in the stack defaults (inherited by all stacks) and override them for specific stacks. See below example config for an example.
 
 ```yaml
 stack_defaults:
   allowed_accounts: '555555555'
 stacks:
   us-east-1:
-    myapp-vpc: # inherits allowed account 555555555
+    myapp-vpc: # only allow account 555555555 (inherited from the stack defaults)
       template: myapp_vpc.rb
       tags:
         purpose: front-end
     myapp-db:
       template: myapp_db.rb
-      allowed_accounts: # only these accounts
+      allowed_accounts: # only allow these accounts (overrides the stack defaults)
         - '1234567890'
         - '9876543210'
       tags:
         purpose: back-end
     myapp-web:
       template: myapp_web.rb
-      allowed_accounts: [] # allow all accounts by overriding stack defaults
+      allowed_accounts: [] # allow all accounts (overrides the stack defaults)
       tags:
         purpose: front-end
     myapp-redis:
       template: myapp_redis.rb
-      allowed_accounts: '888888888' # only this account
+      allowed_accounts: '888888888' # only allow this account (overrides the stack defaults)
       tags:
         purpose: back-end
 ```
--- a/README.md
+++ b/README.md
@@ -91,10 +91,14 @@ stacks:
   staging:
     myapp-vpc:
       template: myapp_vpc.rb
+      allowed_accounts: '123456789'
       tags:
         purpose: front-end
     myapp-db:
       template: myapp_db.rb
+      allowed_accounts:
+        - '1234567890'
+        - '9876543210'
       tags:
         purpose: back-end
     myapp-web:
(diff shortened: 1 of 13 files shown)


lib/stack_master/identity.rb
```diff
@@ -0,0 +1,23 @@
+module StackMaster
+  class Identity
+    def running_in_allowed_account?(allowed_accounts)
```
lukerobins: Seems odd to pass a `nil` or `empty` variable to a function.

stevehodgkiss: Maybe if `nil` was the default value for allowed_accounts (in StackDefinition, indicating unspecified) the expected values would either be `nil` or an array of account IDs

petervandoros: Good idea. Done in b30dede.

lib/stack_master/cli.rb
```diff
@@ -223,15 +230,35 @@ def execute_stacks_command(command, args, options)
...
+        false
+      end
+    end
+
+    def running_in_allowed_account?(allowed_accounts)
+      StackMaster.skip_account_check? || identity.running_in_allowed_account?(allowed_accounts)
+    end
+
+    def identity
+      @account ||= StackMaster::Identity.new
```
viraptor: I think the common idiom is to call the ivar `@identity` (matching the method)

petervandoros: Nice catch! That name was from an earlier iteration of the code. Fixed in 6351741.

README.md
```diff
@@ -545,6 +549,44 @@ end
 
 Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
 
+## Allowed accounts
+
+The AWS account the command is executing in can be restricted to a specific list of allowed accounts. This is useful in reducing the possibility of applying non-production changes in a production account. Each stack definition can specify the `allowed_accounts` property with an array of AWS account IDs the stack is allowed to work with.
+
+This is an opt-in feature which is enabled by specifying at least one account to allow.
+
+Unlike other stack defaults, `allowed_accounts` values specified in the stack definition override values in the stack defaults instead of merging them together. This allows specifying allowed accounts in the stack defaults for all stacks and still have different values for specific stacks. See below example config for an example.
```
patrobinson: From just reading this I'm still not entirely sure what the behaviour will be. Does it allow applying `myapp-db` to account `555555555` or not?

lib/stack_master/identity.rb
```diff
@@ -0,0 +1,23 @@
...
+      @account ||= sts.get_caller_identity.account
+    end
+
+    private
+
+    attr_reader :sts
+
+    def region
+      @region ||= ENV['AWS_REGION'] || Aws.config[:region] || Aws.shared_config.region || 'us-east-1'
+    end
```
patrobinson: This method isn't necessary. If we don't specify a region the SDK does a lookup for us automatically.

spec/stack_master/commands/status_spec.rb
```diff
@@ -35,10 +41,66 @@
...
+        expect { status.perform }.to output(out).to_stdout
+      end
+
+      context 'when --skip-account-check flag is set' do
+        before do
+          StackMaster.skip_account_check!
+        end
+
+        after do
+          StackMaster.reset_flags
```
patrobinson: I think spec_helper already handles this

spec/stack_master/commands/status_spec.rb
```diff
@@ -35,10 +41,66 @@
...
+
+      it 'sets stack status and different fields accordingly' do
+        out = <<~OUTPUT
+          REGION    | STACK_NAME | STACK_STATUS       | DIFFERENT
+          ----------|------------|--------------------|----------
+          us-east-1 | stack1     | Disallowed account | N/A      
+          us-east-1 | stack2     | UPDATE_COMPLETE    | Yes      
+           * No echo parameters can't be diffed
+        OUTPUT
+        expect { status.perform }.to output(out).to_stdout
```
patrobinson: Shouldn't this test expect an error to be returned? If the account IDs don't match it shouldn't print the status?

viraptor: It's a got a special case now: `Disallowed account | N/A`.
It would not be great if it threw an error just because one stack is unavailable.

petervandoros: What @viraptor said.

Alternatively, I could leave them blank instead? I didn't go with that option because I thought it would be better to let the user know why no status is included for that stack.

spec/stack_master/commands/status_spec.rb
petervandoros: It doesn't handle this. The cucumber [features do](https://github.com/envato/stack_master/blob/master/features/support/env.rb#L16) though.

lib/stack_master/identity.rb
petervandoros: You're right it does, however I added this method to ensure that a default region is used if it couldn't find a region from the environment. It's also the same as the CloudFormation driver class except for the default value.

This was me being more defensive, but happy to remove it if you think it'll be fine without it?

spec/stack_master/commands/status_spec.rb
petervandoros: Another option is to include `SKIPPED` for the status field and `Disallowed account` for different field.

README.md
petervandoros: No it doesn't as the account IDs specified in the stack definition overrides the stack defaults.

I've reworded this section and example stack below. Let me know if it clarifies things.

See 4233345.

spec/stack_master/commands/status_spec.rb
patrobinson: Right you are

spec/stack_master/commands/status_spec.rb
patrobinson: Right I see. Are we able to have an integration test for when someone runs `stack_master apply` on a stack in the wrong account then? I thought that's what this was doing, but I didn't read the spec name :laughing:

petervandoros: Good idea. Added integration tests in f02603c.

PR #286, opened 2019-08-14
Enable reading the templates from sparkle-packs

viraptor:
Use a separate compiler option to specify where we should look for the
template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,7 +588,15 @@ end
 
 Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
 
-Templates can be also loaded from sparkle packs by defining `sparkle_pack_template`. No extension should be added in this case:
+Templates can be also loaded from sparkle packs by defining `sparkle_pack_template`. The name corresponds to the registered symbol rather than specific name. That means for a sparkle pack containing:
+
+```ruby
+SparkleFormation.new(:template_name) do
+  ...
+end
+```
+
+we can use stack defined as follows:
 
 ```yaml
 stacks:
--- a/README.md
+++ b/README.md
@@ -588,6 +588,19 @@ end
 
 Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
 
+Templates can be also loaded from sparkle packs by defining the compiler option `sparkle_pack_template`. No extension should be added in this case:
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: template_name
+      compiler_options:
+        sparkle_packs:
+          - some-sparkle-pack
+        sparkle_pack_template: true
+```
+
 ## Allowed accounts
 
 The AWS account the command is executing in can be restricted to a specific list of allowed accounts. This is useful in reducing the possibility of applying non-production changes in a production account. Each stack definition can specify the `allowed_accounts` property with an array of AWS account IDs the stack is allowed to work with.
--- a/README.md
+++ b/README.md
@@ -570,7 +570,7 @@ stacks:
 
 ```yaml
 stacks:
-  us-east-1
+  us-east-1:
     my-stack:
       template: my-stack-with-dynamic.rb
       compiler_options:
@@ -600,7 +600,7 @@ we can use stack defined as follows:
 
 ```yaml
 stacks:
-  us-east-1
+  us-east-1:
     my-stack:
       sparkle_pack_template: template_name
       compiler_options:
--- a/README.md
+++ b/README.md
@@ -588,17 +588,16 @@ end
 
 Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
 
-Templates can be also loaded from sparkle packs by defining the compiler option `sparkle_pack_template`. No extension should be added in this case:
+Templates can be also loaded from sparkle packs by defining `sparkle_pack_template`. No extension should be added in this case:
 
 ```yaml
 stacks:
   us-east-1
     my-stack:
-      template: template_name
+      sparkle_pack_template: template_name
       compiler_options:
         sparkle_packs:
           - some-sparkle-pack
-        sparkle_pack_template: true
 ```
 
 ## Allowed accounts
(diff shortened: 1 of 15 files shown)


viraptor:
Posting this for early review. If the idea is 👌 , I'll add the tests.

README.md
```diff
@@ -588,6 +588,19 @@ end
...
+
+```yaml
+stacks:
+  us-east-1
+    my-stack:
+      template: template_name
+      compiler_options:
+        sparkle_packs:
+          - some-sparkle-pack
+        sparkle_pack_template: true
```
stevehodgkiss: I wonder if it'd be better to omit `template` and specify `sparkle_pack_template: template_name` directly for this scenario 🤔

viraptor: Yup, that's the plan we talked about with Patrick today. Pending change tomorrow.

README.md
```diff
@@ -588,6 +588,18 @@ end
 
 Note though that if a dynamic with the same name exists in your `templates/dynamics/` directory it will get loaded since it has higher precedence.
 
+Templates can be also loaded from sparkle packs by defining `sparkle_pack_template`. No extension should be added in this case:
```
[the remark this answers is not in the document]
viraptor: The idea that the template name is the same as path is just a common way we do it. This doesn't have to be true, and you can have multiple templates per single file. For example you can have a file called `foo.rb` containing:
```
SparkleFormation.new(:foo1) do ... end
SparkleFormation.new(:foo2) do ... end
```
and the exported template names will be `foo1` and `foo2`.

I'll make it explicit in the README, because just referring to the extension is confusing in this case.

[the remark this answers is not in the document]
petervandoros: Thanks for the explanation.

Not sure why someone would want to do that, but I understand that it's possible 👍

README.md
```diff
@@ -588,6 +588,26 @@ end
...
+SparkleFormation.new(:template_name) do
+  ...
+end
+```
+
+we can use stack defined as follows:
+
+```yaml
+stacks:
+  us-east-1
```
petervandoros: Missing `:`
```suggestion
  us-east-1:
```

lib/stack_master/template_compilers/cfndsl.rb
```diff
@@ -4,11 +4,11 @@ def self.require_dependencies
       require 'cfndsl'
     end
 
-    def self.compile(template_file_path, compile_time_parameters, _compiler_options = {})
+    def self.compile(stack_definition, compile_time_parameters, _compiler_options = {})
       CfnDsl.disable_binding
       CfnDsl::ExternalParameters.defaults.clear # Ensure there's no leakage across invocations
       CfnDsl::ExternalParameters.defaults(compile_time_parameters.symbolize_keys)
-      ::CfnDsl.eval_file_with_extras(template_file_path).to_json
+      ::CfnDsl.eval_file_with_extras(stack_definition.template_file_path).to_json
```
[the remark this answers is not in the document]
viraptor: It's needed in the sparkleformation compiler, so all other compilers need comply with the interface. I could expand the API to pass all the needed parameters (so at least `sparkle_pack_template`, `template_file_path`,  and `template_dir`), but that seems excessive.

README.md
```diff
@@ -588,6 +588,26 @@ end
...
+end
+```
+
+we can use stack defined as follows:
+
+```yaml
+stacks:
+  us-east-1:
+    my-stack:
+      sparkle_pack_template: template_name
```
stevehodgkiss: Shall we move this to `compiler_options` (rather than StackDefinition)?

lib/stack_master/template_compiler.rb
```diff
@@ -2,12 +2,12 @@ module StackMaster
   class TemplateCompiler
     TemplateCompilationFailed = Class.new(RuntimeError)
 
-    def self.compile(config, template_file_path, compile_time_parameters, compiler_options = {})
-      compiler = template_compiler_for_file(template_file_path, config)
+    def self.compile(config, template_dir, template_file_path, sparkle_pack_template, compile_time_parameters, compiler_options = {})
```
stevehodgkiss: Having a sparkle specific option on the generic TemplateCompiler isn't ideal... I think it would be better as a compiler_option, or even as it was before using the `template` key and a compiler option.

viraptor: @patrobinson Any opinions? As a user I would expect the template name in the stack options (which I understand as "what to compile") as opposed to compiler option (which I understand as "how to compile"), but I can change my mind :)
I'm still happy to move it around if there's an agreement on a better solution.

patrobinson: @stevehodgkiss I'm confused, I thought you were in support of this change https://github.com/envato/stack_master/pull/286#discussion_r313808613

stevehodgkiss: That comment didn't convey what I intended clearly enough. In my head all the `sparkle_` options would have been compiler options. I don't think compiler specific options should be added directly to the stack definition.

StackMaster supports multiple template compilers - json/yaml/cfndsl/sparkleformation… if we added all their options as arguments to the template compiler (via the stack definition) it would be a mess.

Going forward from here, I'd propose either as it was in this PR before:

```
my-stack:
  template: sparkle_template.rb
  compiler_options:
    sparkle_packs:
      - some-sparkle-pack
    sparkle_pack_template: true
```

Or

```
my-stack:
  compiler_options:
    sparkle_packs:
      - some-sparkle-pack
    sparkle_pack_template: sparkle_template.rb
```

Thinking about it some more, I prefer the first one...

petervandoros: FWIW, I'm a +1 on the first option.

orien: @stevehodgkiss Your argument makes sense to me. Sounds like a good separation of concerns.

I don't know much about the inner-workings of Stack Master to understand the consequences, but from looking at the stack snippet options, from a user's perspective, I like option 1.

patrobinson: The first option was how we had it originally. I wasn't keen on the extra `sparkle_pack_template: true` option because it was a bit superfluous IMO given you could just supply a different parameter but I'm happy either way.

stevehodgkiss: Pushed a PR with those changes and some testing improvements https://github.com/envato/stack_master/pull/289

PR #289, opened 2019-08-16
TemplateCompiler interface / Sparkle config tweaks

stevehodgkiss:
- Move `sparkle_pack_template` out of StackDefinition to compiler_options, with `template` containing the sparkle pack template (as it was originally earlier in the sparkle pack template PR).
- Improve SparkleFormation compiler specs. They were very brittle. Changed them to run SparkleFormation without stubbing it out.
- Changed TemplateCompiler interface to take the template dir and the template (name), instead of the dir and the full path. Although, I wonder if passing `template_file_path` alone would suffice, and use `File.basename` in SparkleFormation compiler to get template dir. It wouldn't work for a template in a subdirectory like `template: vpc/vpc.rb` but this was how it was working before.


stevehodgkiss:
[[SECRETS] StackMaster tweaks around sparkle pack templates](https://trello.com/c/qNoigp9D/748-secrets-stackmaster-tweaks-around-sparkle-pack-templates)

lib/stack_master/template_compiler.rb
```diff
@@ -2,12 +2,16 @@ module StackMaster
   class TemplateCompiler
     TemplateCompilationFailed = Class.new(RuntimeError)
 
-    def self.compile(config, template_dir, template_file_path, sparkle_pack_template, compile_time_parameters, compiler_options = {})
-      compiler = template_compiler_for_stack(template_file_path, sparkle_pack_template, config)
+    def self.compile(config, template_compiler, template_dir, template, compile_time_parameters, compiler_options = {})
+      compiler = if template_compiler
```
viraptor: Is this more readable? (I'm not sure)
```
compiler = @compilers.fetch(template_compiler) { |_| template_compiler_for_stack(template, config) }
```

spec/stack_master/template_compilers/sparkle_formation_spec.rb
```diff
@@ -1,168 +1,119 @@
...
+      end
 
-    it 'should validate the parameters against any compile time definitions' do
-      expect(StackMaster::SparkleFormation::CompileTime::ParametersValidator).to receive(:new).with(compile_time_parameter_definitions, compile_time_parameters)
-      expect(parameters_validator).to receive(:validate)
-      compile
-    end
+      context 'compile time parameters validations' do
+        it 'should validate the compile time definitions' do
+          definitions_validator = double
```
viraptor: `instance_double`?

lib/stack_master/template_compiler.rb
stevehodgkiss: Hmmm, I think I'd rather not look in `@compilers` if `template_compiler` is nil 🤔

README.md
```diff
@@ -602,10 +602,12 @@ we can use stack defined as follows:
 stacks:
   us-east-1:
     my-stack:
-      sparkle_pack_template: template_name
+      template: template_name
```
petervandoros: Just wanting to confirm my understanding is correct and that `template` includes the template's file extension?

README.md
```diff
@@ -602,10 +602,12 @@ we can use stack defined as follows:
...
+      compiler: sparkle_formation
```
petervandoros: Did you consider adding the compiler type to the `compiler_options` properties? If so, I'm curious to understand why you chose to not to.

README.md
stevehodgkiss: It usually does, but for sparkle pack templates it refers to the symbol key used in `SparkleFormation.new(:template_name) do`

README.md
stevehodgkiss: Yeah I briefly considered that. I went with a specific attribute on StackDefinition because it's a value for StackMaster, not the compiler itself.

lib/stack_master/template_compiler.rb
```diff
@@ -16,12 +16,8 @@ def self.register(name, klass)
...
     # private
-    def self.template_compiler_for_stack(template_file_path, sparkle_pack_template, config)
-      ext = if sparkle_pack_template
-        :rb
-      else
-        file_ext(template_file_path)
-      end
+    def self.template_compiler_for_stack(template_file_path, config)
+      ext = file_ext(template_file_path)
       compiler_name = config.template_compilers.fetch(ext)
```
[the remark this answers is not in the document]
stevehodgkiss: It's necessary to specify `compiler` for sparkle pack templates. Using symbol to lookup because it's hard to lookup the template by file name, and there could technically be multiple templates defined in a single file.

README.md
petervandoros: That's unfortunate 😞

PR #290, opened 2019-09-24
Add a Troubleshooting section and add service-linked roles

mattpatterson94:
Recently I had an issue when applying a stack. This is due to our aws account not having the necessary service-linked role allowing aws to make changes from one service to another.

I have started a Troubleshooting guide here to feature things like this. This seems like the best place to put it because it will usually happen when you're using stack_master.


mattpatterson94:
This isn't the right place to put this info

PR #292, opened 2019-11-12
Cross-account parameter resolving

petervandoros:
This PR adds support for assuming a role when resolving parameters.

We've recently had to configure VPC peering and found it error prone to have to hard code VPC and peering IDs in the stacks' parameters files. It's also hard and tedious to find all the relevant values, especially for old VPCs.

Being able to reference stack outputs for stacks in other accounts would make this much more maintainable and less error prone, especially if the stacks need re-creating.

Here's some examples of what it looks like:

```yaml
vpc_peering_id:
  role: cross-account-parameter-resolver
  account: 1234567890
  stack_output: vpc-peering-stack-in-other-account/peering_id
an_array_param:
  role: cross-account-parameter-resolver
  account: 1234567890
  stack_outputs:
    - stack-in-account1/output
    - stack-in-account1/another_output
another_array_param:
  - role: cross-account-parameter-resolver
    account: 1234567890
    stack_output: stack-in-account1/output
  - role: cross-account-parameter-resolver
    account: 0987654321
    stack_output: stack-in-account2/output
```

#### Considerations

- The relevant IAM permissions to allow cross-account parameter resolving is left as an exercise for the reader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,10 @@ region/account, even though the resolved values will be different.
 One way to resolve parameter values from different accounts to the one StackMaster runs in, is to
 assume a role in another account with the relevant IAM permissions to execute successfully.
 
-This is supported in StackMaster by specifying the `role` and `account` properties for any
+This is supported in StackMaster by specifying the `role` and `account` properties for the
 parameter resolver in the stack's parameters file.
+
+All parameter resolvers are supported.
 
 ```yaml
 vpc_peering_id:
@@ -226,6 +228,11 @@ another_array_param:
   - role: cross-account-parameter-resolver
     account: 0987654321
     stack_output: stack-in-account2/output
+
+my_secret:
+  role: cross-account-parameter-resolver
+  account: 1234567890
+  parameter_store: ssm_parameter_name
 ```
 
 An example of use case where cross-account parameter resolving is particularly useful is when
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ db_password:
 ```
 
 ### 1Password Lookup
-An Alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
+An alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
 You declare a 1password lookup with the following parameters in your parameters file:
 
 ```
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ db_password:
 ```
 
 ### 1Password Lookup
-An alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
+An alternative to the secrets store is accessing 1password secrets using the 1password cli (`op`).
 You declare a 1password lookup with the following parameters in your parameters file:
 
 ```
--- a/README.md
+++ b/README.md
@@ -198,6 +198,40 @@ One benefit of using parameter resolvers instead of hard coding values like VPC
 IDs and resource ARNs is that the same configuration works cross
 region/account, even though the resolved values will be different.
 
+### Cross-account parameter resolving
+
+One way to resolve parameter values from different accounts to the one StackMaster runs in, is to
+assume a role in another account with the relevant IAM permissions to execute successfully.
+
+This is supported in StackMaster by specifying the `role` and `account` properties for any
+parameter resolver in the stack's parameters file.
+
+```yaml
+vpc_peering_id:
+  role: cross-account-parameter-resolver
+  account: 1234567890
+  stack_output: vpc-peering-stack-in-other-account/peering_name
+
+an_array_param:
+  role: cross-account-parameter-resolver
+  account: 1234567890
+  stack_outputs:
+    - stack-in-account1/output
+    - stack-in-account1/another_output
+
+another_array_param:
+  - role: cross-account-parameter-resolver
+    account: 1234567890
+    stack_output: stack-in-account1/output
+  - role: cross-account-parameter-resolver
+    account: 0987654321
+    stack_output: stack-in-account2/output
+```
+
+An example of use case where cross-account parameter resolving is particularly useful is when
+setting up VPC peering where you need the VPC ID of the peer. Without the ability to assume
+a role in another account, the only option was to hard code the peer's VPC ID.
+
 ### Stack Output
 
 The stack output parameter resolver looks up outputs from other stacks in the
--- a/features/apply_with_assume_role_parameter_resolvers.feature
+++ b/features/apply_with_assume_role_parameter_resolvers.feature
@@ -1,0 +1,57 @@
+Feature: Apply command with assume role parameter resolvers
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us-east-2:
+          vpc:
+            template: vpc.rb
+          myapp_web:
+            template: myapp_web.rb
+      """
+    And a directory named "parameters"
+    And a file named "parameters/myapp_web.yml" with:
+      """
+      vpc_id:
+        role: my-role
+        account: 1234567890
+        stack_output: vpc/vpc_id
+      """
+    And a directory named "templates"
+    And a file named "templates/myapp_web.rb" with:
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -198,6 +198,40 @@ One benefit of using parameter resolvers instead of hard coding values like VPC
...
+assume a role in another account with the relevant IAM permissions to execute successfully.
+
+This is supported in StackMaster by specifying the `role` and `account` properties for any
+parameter resolver in the stack's parameters file.
+
+```yaml
+vpc_peering_id:
+  role: cross-account-parameter-resolver
+  account: 1234567890
+  stack_output: vpc-peering-stack-in-other-account/peering_name
```
orien: It looks like this'd work with other parameter resolvers. If so should we add some examples in? SSM parameter store for instance:

```yaml
my_secret:
  role: cross-account-parameter-resolver
  account: 1234567890
  parameter_store: ssm_parameter_name
```

lib/stack_master/role_assumer.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def assume_role_credentials(account, role)
+      credentials_key = "#{account}:#{role}"
+      @credentials.fetch(credentials_key) do
+        @credentials[credentials_key] = Aws::AssumeRoleCredentials.new(
+          role_arn: "arn:aws:iam::#{account}:role/#{role}",
+          role_session_name: "stack-master-assume-role-parameter-resolver"
+        )
+      end
```
patrobinson: Why do we want to save credentials to an instance variable?

spec/stack_master/parameter_resolver_spec.rb
```diff
@@ -98,6 +98,91 @@ def resolve(params)
...
+            'my_resolver' => 2
+          }
+        }
+      end
+
+      it 'caches the role assumer' do
+        expect(StackMaster::RoleAssumer).to receive(:new).once
+
+        parameter_resolver.resolve
+      end
```
patrobinson: right I see now

lib/stack_master/role_assumer.rb
petervandoros: So it doesn't have to assume the role for every parameter wishing to assume the same role.

README.md
petervandoros: Good idea.

I also tweaked the wording a little to be explicit about which parameter resolvers support assuming a role in 31bbaa3. I.e., all of them!

lib/stack_master/role_assumer.rb
```diff
@@ -0,0 +1,59 @@
...
+        block.call
+      ensure
+        restore_aws_global_config(original_aws_config)
+        restore_cf_driver(original_cf_driver)
+      end
+    end
+
+    private
+
+    def replace_aws_global_config
```
viraptor: I think the replace/restore functions are not very pretty. Part of that logic also exists in `assume_role`.

How about:

```ruby
def assume_role(...)
...
  role_credentials = assume_role_credentials(account, role)
  with_temporary_role(role_credentials) do
    with_temporary_driver do
      block.call
    end
  end
...
end

def with_temporary_role(credentials, &block)
  original_aws_config = Aws.config
  Aws.config = original_aws_config.deep_dup
  block.call
ensure
  Aws.config = original_aws_config
end

def with_temporary_driver(&block)
  original_driver = StackMaster.cloud_formation_driver
  StackMaster.cloud_formation_driver = original_driver.class.new
  block.call
ensure
  StackMaster.cloud_formation_driver = original_driver
end
```

lib/stack_master/role_assumer.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def replace_cf_driver
+      driver = StackMaster.cloud_formation_driver
+      StackMaster.cloud_formation_driver = driver.class.new
+      driver
+    end
+
+    def restore_cf_driver(driver)
+      return if driver.nil?
```
viraptor: If the driver is nil, `driver.class.new` in `replace_cf_driver` would fail earier.

orien:
Do you mind adding to the [changelog](https://github.com/envato/stack_master/blob/master/CHANGELOG.md)?

lib/stack_master/role_assumer.rb
petervandoros: Great idea. Done in 95a8825.

petervandoros:
@orien Updated changelog in 41592b3.

petervandoros:
@orien @stevehodgkiss @patrobinson I've made a few more changes to resolve an issue I found.

lib/stack_master/parameter_resolvers/ejson.rb
```diff
@@ -27,9 +28,12 @@ def validate_ejson_file_specified
       end
 
       def decrypt_ejson_file
-        @decrypt_ejson_file ||= EJSONWrapper.decrypt(ejson_file_path,
-                                                     use_kms: @stack_definition.ejson_file_kms,
-                                                     region: ejson_file_region)
+        ejson_file_key = credentials_key
```
petervandoros: @stevehodgkiss Is my assumption that the scope for the cached decrypted secrets is the credentials?

README.md
```diff
@@ -270,7 +311,7 @@ db_password:
 ```
 
 ### 1Password Lookup
-An Alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
+An alternative to the alternative secret store is accessing 1password secrets using the 1password cli (`op`).
```
stevehodgkiss: This sentence could use some work 😂

lib/stack_master/parameter_resolvers/ejson.rb
stevehodgkiss: Yeah, I think that assumption is correct. The file can't change between parameters right now (it's in stack definition / stack_master.yml config).

lib/stack_master/role_assumer.rb
```diff
@@ -0,0 +1,54 @@
...
+      StackMaster.cloud_formation_driver = new_driver
+      block.call
+    ensure
+      StackMaster.cloud_formation_driver = original_driver
+    end
+
+    def assume_role_credentials(account, role)
+      credentials_key = "#{account}:#{role}"
+      @credentials.fetch(credentials_key) do
+        @credentials[credentials_key] = Aws::AssumeRoleCredentials.new(
```
stevehodgkiss: Is it necessary to set the hash here if fetch is being used?

```suggestion
        Aws::AssumeRoleCredentials.new(
```

features/apply_with_assume_role_parameter_resolvers.feature
```diff
@@ -0,0 +1,57 @@
+Feature: Apply command with assume role parameter resolvers
```
stevehodgkiss: ❤️

README.md
petervandoros: Not sure how that change got in there :/

Fixed in 4ad4ade

lib/stack_master/parameter_resolvers/ejson.rb
petervandoros: That's my understanding too. Thanks for confirming.

lib/stack_master/role_assumer.rb
petervandoros: The block passed to the `fetch` method specifies the return value if the key is not found. It doesn't set the value in the hash if it's missing. We need to set the hash value for it to persist.

```
irb(main):001:0> h = {a: 'a', b: 'b'}
=> {:a=>"a", :b=>"b"}
irb(main):002:0> h.fetch(:z) do
irb(main):003:1* 'z'
irb(main):004:1> end
=> "z"
irb(main):005:0> h
=> {:a=>"a", :b=>"b"}
irb(main):006:0> h.fetch(:z) do
irb(main):007:1* h[:z] = 'z'
irb(main):008:1> end
=> "z"
irb(main):009:0> h
=> {:a=>"a", :b=>"b", :z=>"z"}
```

PR #294, opened 2019-12-16
Fix TravisCI errors

petervandoros:
Travis CI build have started [failing](https://travis-ci.org/envato/stack_master/builds/625954645). This PR aims to resolve those build failures.


stack_master.gemspec
```diff
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ["lib"]
-  spec.required_ruby_version = ">= 2.1.0"
+  spec.required_ruby_version = ">= 2.4.0"
```
orien: I wonder about this. It explicitly breaks compatibility. Do we need a major version bump to release this change?

petervandoros: We don't need a major version bump to release this change. Shouldn't we keep this in sync with what we support?

petervandoros: I don't feel strongly about this so I reverted this change in a53e340.

orien: > We don't need a major version bump to release this change.

I feel this is in a grey area and is not so clear cut. I'd like to hear more opinions on it. If version 1.17.1 runs on Ruby 2.3, and version 1.18.0 does not, is that a change to the public API? What is the expectation in the Ruby community?

Just because we're not testing a specific version, doesn't mean we should necessarily ban it. For example, we have the dependency `'activesupport', '>= 4'`. This means we support version 4 and up. But the test suite only tests against the latest version (v6.0.1 at the moment).

PR #295, opened 2020-01-13
Remove GPG secret parameter resolver

orien:
Please add `stack_master-gpg_parameter_resolver` to your bundle to continue using the GPG parameter resolving functionality.


patrobinson:
@orien there's a few references still in the README. Can we replace these with links to the gpg gem?

orien:
@patrobinson moved in b33168b and https://github.com/envato/stack_master-gpg_parameter_resolver/pull/1

README.md
```diff
@@ -136,7 +135,7 @@ stacks:
 
 - `templates` - CloudFormation, SparkleFormation or CfnDsl templates.
 - `parameters` - Parameters as YAML files.
-- `secrets` - GPG encrypted secret files.
+- `secrets` - encrypted secret files.
```
patrobinson: I don't think this parameter makes sense anymore without the GPG resolver... It's an unfortunate artefact of how it was originally written that we need to keep it. Perhaps we should just specify this option is only used for the GPG resolver.

README.md
```diff
@@ -264,35 +261,10 @@ into parameters of dependent stacks.
 
 ### Secret
 
-Note: This resolver is not supported on Windows, you can instead use the [Parameter Store](#parameter-store).
+Note: This resolver has been extracted into a dedicated gem. Please install and
```
patrobinson:
```suggestion
Note: The GPG parameter resolver has been extracted into a dedicated gem. Please install and
```

lib/stack_master/stack_definition.rb
```diff
@@ -9,7 +9,6 @@ class StackDefinition
                   :notification_arns,
                   :base_dir,
                   :template_dir,
-                  :secret_file,
```
patrobinson: The stack definition still needs to contain this because the GPG resolver depends on it https://github.com/envato/stack_master-gpg_parameter_resolver/blob/master/lib/stack_master/parameter_resolvers/secret.rb#L43

README.md
orien: This is actually a list of directories used by StackMaster. The `secrets` directory is still used by the EJSON parameter resolver.

lib/stack_master/stack_definition.rb
orien: The GPG Parameter Resolver actually adds this back in.

https://github.com/envato/stack_master-gpg_parameter_resolver/blob/5eacf9639941aa830ef0f640ef9a5afb7e76eaf5/lib/stack_master/parameter_resolvers/secret.rb#L4-L6

This is a rather dubious approach though. Thoughts?

orien: Actually, I'm preferring to move all the support into the gem. That way it doesn't need to be supported in StackMaster, we can make a clean break in version 2.

Later we could re-work the GPG Parameter Resolver to not use this configuration.

patrobinson: Wholly monkey patching that's ugly.

Ok let's leave it as is and refactor for a v1 of GPG parameter resolver.

PR #303, opened 2020-02-13
Add support to set AWS profile

sideshowcecil:
Add support to switch the AWS profile via environment variable `AWS_PROFILE` like in the `aws` CLI.


sideshowcecil:
Well indeed it is, I was having some issues on my end and blamed stack_master for it.
So let's close this one :)

PR #305, opened 2020-02-16
Add a "tidy" command to find files that might be missing or unneeded.

mipearson:
Sometimes you delete stacks but don't remove their parameter files or templates.

Sometimes a template is renamed or removed, but stacks still depend on it.

`stack_master tidy` (I feel like it needs a better name) helps you find these things.

### DIFF
--- a/lib/stack_master/commands/tidy.rb
+++ b/lib/stack_master/commands/tidy.rb
@@ -37,7 +37,7 @@ module StackMaster
         end
       end
 
-      def rel_path path
+      def rel_path(path)
         Pathname.new(path).relative_path_from(Pathname.new(@config.base_dir))
       end
 
--- a/lib/stack_master/commands/tidy.rb
+++ b/lib/stack_master/commands/tidy.rb
@@ -1,0 +1,68 @@
+module StackMaster
+  module Commands
+    class Tidy
+      include Command
+      include StackMaster::Commands::TerminalHelper
+
+      def initialize(config)
+        @config = config
+      end
+
+      def perform
+        used_templates = []
+        used_parameter_files = []
+
+        templates = Set.new(find_templates())
+        parameter_files = Set.new(find_parameter_files())
+
+        status = @config.stacks.each do |stack_definition|
+          parameter_files.subtract(stack_definition.parameter_files)
+          template = File.absolute_path(stack_definition.template_file_path)
+
+          if template
+            templates.delete(template)
+
+            if !File.exist?(template)
+              StackMaster.stdout.puts "Stack \"#{stack_definition.stack_name}\" in \"#{stack_definition.region}\" missing template \"#{rel_path(template)}\""
+            end
+          end
+        end
+
+        templates.each do |path|
+          StackMaster.stdout.puts "#{rel_path(path)}: no stack found for this template"
+        end
+
+        parameter_files.each do |path|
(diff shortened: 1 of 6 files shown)


lib/stack_master/commands/tidy.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def perform
+        used_templates = []
+        used_parameter_files = []
+
+        templates = Set.new(find_templates())
+        parameter_files = Set.new(find_parameter_files())
+
+        status = @config.stacks.each do |stack_definition|
```
viraptor: I feel like this could be a separate method like `def remove_used(templates, parameter_files)` for clarity. (or maybe even one for each?) But that's just nitpicking :)

lib/stack_master/commands/tidy.rb
```diff
@@ -0,0 +1,68 @@
...
+        templates.each do |path|
+          StackMaster.stdout.puts "#{rel_path(path)}: no stack found for this template"
+        end
+
+        parameter_files.each do |path|
+          StackMaster.stdout.puts "#{rel_path(path)}: no stack found for this parameter file"
+        end
+      end
+
+      def rel_path path
```
viraptor: I think the rest of the code uses parens around parameters.

lib/stack_master/commands/tidy.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+
+        templates.each do |path|
+          StackMaster.stdout.puts "#{rel_path(path)}: no stack found for this template"
+        end
+
+        parameter_files.each do |path|
+          StackMaster.stdout.puts "#{rel_path(path)}: no stack found for this parameter file"
+        end
+      end
```
viraptor: Can we return a non-zero exit code if any of them is not empty? That would be nice for CI.

mipearson: Thought about it, but they're mostly indications rather than errors - eg in the data platform repositories, there's multiple stack_master.yml files that share a template directory.

They'll both find "extra" files as some templates are only used by one of the two configs .. but they're exceptions rather than the rule.

lib/stack_master/commands/tidy.rb
mipearson: Had a go at this but as it's referencing multiple sets (and has side effects where it prints missing templates) it complicates things a bit. Decided it's not worth it for this length of method.

lib/stack_master/commands/tidy.rb
patrobinson: I wonder if we should have a flag to enable or disable checking for orphaned templates?

mipearson: Maybe - my intention for this wasn't as a CI pass/fail tool but a convenience one, especially as the method for knowing which directories/files are referenced is not 100% reliable.

lib/stack_master/commands/tidy.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def perform
+        used_templates = []
+        used_parameter_files = []
+
+        templates = Set.new(find_templates())
+        parameter_files = Set.new(find_parameter_files())
+
+        status = @config.stacks.each do |stack_definition|
+          parameter_files.subtract(stack_definition.parameter_files)
```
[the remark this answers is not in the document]
patrobinson: Oh nevermind,I understand what's going on here now. `parameter files - (stacks)`

PR #310, opened 2020-03-07
Handle exit status from the CLI class

### LINKED ISSUE #183: stack_master --version returns a failed exit code
Is this intended?  Not critical, but a little bizarre.

```
arecker@localhost:testproj$ be stack_master --version
StackMaster 0.16.0
arecker@localhost:testproj$ echo $?
1
```

orien:
### Context

There are a couple of peculiar StackMaster behaviours regarding exit status.

- `stack_master --version` results in a status of `1` (#183).
- The exit status is set in the `stack_master` binstub, but the Cucumber feature suite doesn't test this. It drives StackMaster via the `StackMaster::CLI` class directly. This makes validating the status code impossible.

### Change

I propose we move handling of the exit status into the `StackMaster::CLI` class. This will solve both problems:

- Running `stack_master --version` will return a status of `0` (fixes #183).
- The Cucumber test suite will accurately validate the exit status of each command. I've added this validation to all scenarios.

### DIFF
--- a/bin/stack_master
+++ b/bin/stack_master
@@ -10,8 +10,7 @@ end
 trap("SIGINT") { raise StackMaster::CtrlC }
 
 begin
-  result = StackMaster::CLI.new(ARGV.dup).execute!
-  exit !!result
+  StackMaster::CLI.new(ARGV.dup).execute!
 rescue StackMaster::CtrlC
   StackMaster.stdout.puts "Exiting..."
 end
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -123,6 +123,7 @@ Feature: Apply command
       | +    "Vpc": {                                                                  |
       | Parameters diff:                                                               |
       | KeyName: my-key                                                                |
+    And the exit status should be 0
 
   Scenario: Run apply nothing and create 2 stacks
     Given I stub the following stack events:
@@ -147,6 +148,7 @@ Feature: Apply command
   Scenario: Run apply with invalid stack
     When I run `stack_master apply foo bar`
     Then the output should contain "Could not find stack definition bar in region foo"
+    And the exit status should be 1
 
   Scenario: Create stack with --changed
     Given I stub the following stack events:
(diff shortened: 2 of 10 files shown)


bin/stack_master
```diff
@@ -10,8 +10,7 @@ end
 trap("SIGINT") { raise StackMaster::CtrlC }
 
 begin
-  result = StackMaster::CLI.new(ARGV.dup).execute!
-  exit !!result
+  StackMaster::CLI.new(ARGV.dup).execute!
```
orien: This binstub no longer handles the exit status.

lib/stack_master/cli.rb
```diff
@@ -253,7 +253,7 @@ def execute_stacks_command(command, args, options)
           end
         end
       end
-      success
+      @kernel.exit false unless success
```
orien: The `CLI` class now handles the exit status.

features/version.feature
```diff
@@ -0,0 +1,4 @@
+Feature: Check the StackMaster version
+  Scenario: Use the --version option
+    When I run `stack_master --version`
+    Then the exit status should be 0
```
orien: Add a scenario to ensure `stack_master --version` results in a `0` status.

features/apply.feature
```diff
@@ -123,6 +123,7 @@ Feature: Apply command
       | +    "Vpc": {                                                                  |
       | Parameters diff:                                                               |
       | KeyName: my-key                                                                |
+    And the exit status should be 0
```
orien: Validate the exit status in all Cucumber scenarios.

features/apply_with_allowed_accounts.feature
```diff
@@ -43,7 +43,7 @@ Feature: Apply command with allowed accounts
     And I run `stack_master apply us-east-1 myapp-db`
     And the output should contain all of these lines:
       | Account '11111111' is not an allowed account. Allowed accounts are ["22222222"].|
-    Then the exit status should be 0
+    Then the exit status should be 1
```
orien: This is an example of an incorrect Cucumber validation. This scenario was ensuring a status `0`, but because it was only testing via the `CLI` class it missed the binstub returning `1`.

Now the feature correctly checks the value return in the command invocation.

Note: this is not a change in StackMaster behaviour.

PR #313, opened 2020-03-12
Exit status 1 if stack is not found in AWS

orien:
As reported in https://github.com/envato/stack_master/issues/235#issuecomment-393512594, StackMaster aborts, but with exit status `0` upon not finding the specified stack in AWS. It is expected to report an error via a non-zero exit status.

Update the `delete`, `outputs`, and `resources` commands to exit with a status of `1` if the specified stack does not exist in AWS.

Also the `delete` command now exits with status `1` if using a disallowed AWS account.

### DIFF
--- a/features/delete.feature
+++ b/features/delete.feature
@@ -26,15 +26,31 @@ Feature: Delete command
     When I run `stack_master delete us-east-1 myapp-vpc --trace`
     And the output should contain all of these lines:
       | Stack does not exist |
-    Then the exit status should be 0
+    Then the exit status should be 1
 
   Scenario: Answer no when asked to delete stack
     Given I will answer prompts with "n"
     And I stub the following stacks:
       | stack_id | stack_name | parameters       | region    |
-      | 1        | myapp-vpc  | KeyName=my-key   | us-east-1 | 
+      | 1        | myapp-vpc  | KeyName=my-key   | us-east-1 |
     When I run `stack_master delete us-east-1 myapp-vpc --trace`
     And the output should contain all of these lines:
       | Stack update aborted |
     Then the exit status should be 0
 
+  Scenario: Run a delete command on a stack with the wrong account
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us_east_1:
+          myapp:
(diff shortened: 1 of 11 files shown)


lib/stack_master/cli.rb
```diff
@@ -206,10 +206,11 @@ def execute!
             region = args[0]
           end
 
-          execute_if_allowed_account(allowed_accounts) do
+          success = execute_if_allowed_account(allowed_accounts) do
             StackMaster.cloud_formation_driver.set_region(region)
-            StackMaster::Commands::Delete.perform(region, stack_name)
+            StackMaster::Commands::Delete.perform(region, stack_name).success?
           end
+          @kernel.exit false unless success
```
orien: `stack_master delete` is updated to exit with status `1` if the command object is not successful, or a disallowed AWS account is used.

features/delete.feature
```diff
@@ -26,15 +26,31 @@ Feature: Delete command
...
+            template: myapp.rb
+            allowed_accounts: '11111111'
+      """
+    When I use the account "33333333"
+    And I run `stack_master delete us-east-1 myapp`
+    Then the output should contain:
+      """
+      Account '33333333' is not an allowed account. Allowed accounts are ["11111111"].
+      """
+    And the exit status should be 1
```
orien: Add a feature, specifying the behaviour of the `delete` command when using a disallowed AWS account.

lib/stack_master/commands/resources.rb
```diff
@@ -13,14 +15,14 @@ def perform
...
-          StackMaster.stdout.puts "Stack doesn't exist"
+          failed("Stack doesn't exist")
         end
       end
 
       private
 
       def stack_resources
-        @stack_resources = cf.describe_stack_resources(stack_name: @stack_definition.stack_name).stack_resources
+        @stack_resources ||= cf.describe_stack_resources(stack_name: @stack_definition.stack_name).stack_resources
```
orien: Memoising this results in less API calls.

orien:
☝️ Just rebased against master to resolve the merge conflicts.

PR #316, opened 2020-03-13
Apply command prints list of parameter file locations

orien:
#### Context

I found that if a parameter file couldn't be found for a stack, StackMaster would print this curious line:

```
Empty/blank parameters detected, ensure values exist for those parameters. Parameters will be read from the following locations:
```

But no locations were presented.

Upon investigation, I found it's because we're using `Dir.glob`. This only returns files actually present on the file system.

https://github.com/envato/stack_master/blob/28e088414ad85a13176bc901136e405cb45b3c5a/lib/stack_master/stack_definition.rb#L113-L115

#### Change

I propose to fix this by introducing a way to obtain the glob patterns before they're provided to `Dir.glob`.

With this change it'll print something like:

```
Empty/blank parameters detected, ensure values exist for those parameters.
Parameters will be read from the following locations:
- parameters/myapp.y*ml
- parameters/us-east-1/myapp.y*ml
- parameters/production/myapp.y*ml
```

### DIFF
--- a/lib/stack_master/stack_definition.rb
+++ b/lib/stack_master/stack_definition.rb
@@ -23,7 +23,6 @@ module StackMaster
     include Utils::Initializable
 
     def initialize(attributes = {})
-      @additional_parameter_lookup_dirs = []
       @compiler_options = {}
       @notification_arns = []
       @s3 = {}
@@ -32,6 +31,7 @@ module StackMaster
       @ejson_file_kms = true
       @compiler = nil
       super
+      @additional_parameter_lookup_dirs ||= []
       @template_dir ||= File.join(@base_dir, 'templates')
       @allowed_accounts = Array(@allowed_accounts)
     end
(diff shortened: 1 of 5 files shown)


lib/stack_master/stack_definition.rb
```diff
@@ -32,6 +31,7 @@ def initialize(attributes = {})
       @ejson_file_kms = true
       @compiler = nil
       super
+      @additional_parameter_lookup_dirs ||= []
```
orien: `super` sets `@additional_parameter_lookup_dirs` to `nil` when not provided in the `stack_master.yml`. Moving this below, means we get a sensible default values of `[]`.

lib/stack_master/stack_definition.rb
```diff
@@ -86,7 +86,11 @@ def s3_template_file_name
     end
 
     def parameter_files
-      [ default_parameter_file_path, region_parameter_file_path, additional_parameter_lookup_file_paths ].flatten.compact
+      parameter_file_globs.map(&Dir.method(:glob)).flatten
+    end
+
+    def parameter_file_globs
+      [ default_parameter_glob, region_parameter_glob ] + additional_parameter_lookup_globs
     end
```
orien: This is the new method that provides the parameter file glob patterns.

PR #317, opened 2020-03-13
`stack_master apply` with missing params exits with status 1

orien:
As reported in #235, `stack_master apply` aborts when parameters are missing, however the exit status is `0` reporting success.

I propose to change this so that the exit status is `1` reporting failure.

### DIFF
--- a/lib/stack_master/commands/apply.rb
+++ b/lib/stack_master/commands/apply.rb
@@ -210,16 +210,16 @@ module StackMaster
 
       def ensure_valid_parameters!
         if @proposed_stack.missing_parameters?
-          StackMaster.stderr.puts <<~MESSAGE
+          message = <<~MESSAGE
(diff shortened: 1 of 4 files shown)


lib/stack_master/commands/apply.rb
```diff
@@ -210,16 +210,16 @@ def execute_change_set
...
             Parameters will be read from the following locations:
           MESSAGE
           base_dir = Pathname.new(@stack_definition.base_dir)
           @stack_definition.parameter_file_globs.each do |glob|
             parameter_file = Pathname.new(glob).relative_path_from(base_dir)
-            StackMaster.stderr.puts " - #{parameter_file}"
+            message << " - #{parameter_file}\n"
           end
-          halt!
+          failed!(message)
```
orien: Using `failed!` command method is all that's needed to halt the process and exit with status `1`.

PR #318, opened 2020-03-16
Fewer ActiveSupport core extensions

orien:
Require fewer ActiveSupport core extensions.

### DIFF
--- a/bin/stack_master
+++ b/bin/stack_master
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 
-require 'rubygems'
 require 'stack_master'
 
 if ENV['STUB_AWS'] == 'true'
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -8,7 +8,9 @@ require 'aws-sdk-s3'
 require 'aws-sdk-sns'
 require 'aws-sdk-ssm'
 require 'colorize'
-require 'active_support/core_ext/string'
+require 'active_support/core_ext/hash/keys'
+require 'active_support/core_ext/object/blank'
+require 'active_support/core_ext/string/inflections'
 require 'multi_json'
 
 MultiJson.use :json_gem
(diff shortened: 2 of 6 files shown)


lib/stack_master.rb
```diff
@@ -8,7 +8,9 @@
 require 'aws-sdk-sns'
 require 'aws-sdk-ssm'
 require 'colorize'
-require 'active_support/core_ext/string'
+require 'active_support/core_ext/hash/keys'
+require 'active_support/core_ext/object/blank'
+require 'active_support/core_ext/string/inflections'
```
orien: Previously, when requiring `active_support/core_ext/string` it'd load all these string extensions:

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/string.rb

Which would in turn load a bunch of other core extensions.

Let's be a little more explicit about the extensions StackMaster uses, and only load these.

lib/stack_master/parameter_resolvers/latest_ami.rb
```diff
@@ -12,7 +12,7 @@ def resolve(value)
         owners = Array(value.fetch('owners', 'self').to_s)
         ami_finder = AmiFinder.new(@stack_definition.region)
         filters = ami_finder.build_filters_from_hash(value.fetch('filters'))
-        ami_finder.find_latest_ami(filters, owners).try(:image_id)
+        ami_finder.find_latest_ami(filters, owners)&.image_id
```
orien: Now that we require Ruby >= 2.4, we can use the Ruby safe navigation operator, instead of ActiveSupport's `try`.

lib/stack_master/sparkle_formation/template_file.rb
```diff
@@ -71,7 +71,7 @@ def format
             newlines = lines.split("\n").map do |line|
               "#{line}#{newlines.pop}"
             end
-            if lines.starts_with?("\n")
+            if lines.start_with?("\n")
```
orien: Active support aliases Ruby's `start_with?` to `starts_with?`.

Let's just use the version that comes with Ruby.

bin/stack_master
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 
-require 'rubygems'
```
orien: This is not needed.

PR #319, opened 2020-03-16
Print error backtrace when provided --trace option

orien:
#### Context

StackMaster commands rescue a couple of error types during processing.

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/command.rb#L25-L27

However, it's difficult to get more information on what went wrong as only the class and message is printed to the terminal.

#### Change

Also print the backtrace if the `--trace` option is provided.

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -178,7 +178,7 @@ module StackMaster
           options.defaults config: default_config_file
           say "Invalid arguments. stack_master tidy" and return unless args.size == 0
           config = load_config(options.config)
-          StackMaster::Commands::Tidy.perform(config, options)
+          StackMaster::Commands::Tidy.perform(config, nil, options)
         end
       end
 
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -71,7 +71,7 @@ module StackMaster
           unless args.size == 2
             say "Invalid arguments. stack_master init [region] [stack_name]"
           else
-            StackMaster::Commands::Init.perform(options.overwrite, *args)
+            StackMaster::Commands::Init.perform(options, *args)
           end
         end
       end
@@ -119,7 +119,7 @@ module StackMaster
           options.defaults config: default_config_file
           say "Invalid arguments." if args.size > 0
           config = load_config(options.config)
-          StackMaster::Commands::ListStacks.perform(config)
+          StackMaster::Commands::ListStacks.perform(config, nil, options)
         end
       end
 
@@ -165,7 +165,7 @@ module StackMaster
           options.defaults config: default_config_file
           say "Invalid arguments. stack_master status" and return unless args.size == 0
           config = load_config(options.config)
-          StackMaster::Commands::Status.perform(config)
+          StackMaster::Commands::Status.perform(config, nil, options)
         end
       end
 
@@ -178,7 +178,7 @@ module StackMaster
           options.defaults config: default_config_file
           say "Invalid arguments. stack_master tidy" and return unless args.size == 0
           config = load_config(options.config)
-          StackMaster::Commands::Tidy.perform(config)
+          StackMaster::Commands::Tidy.perform(config, options)
         end
       end
 
@@ -208,7 +208,7 @@ module StackMaster
 
           success = execute_if_allowed_account(allowed_accounts) do
             StackMaster.cloud_formation_driver.set_region(region)
-            StackMaster::Commands::Delete.perform(region, stack_name).success?
+            StackMaster::Commands::Delete.perform(region, stack_name, options).success?
           end
           @kernel.exit false unless success
         end
--- a/lib/stack_master/command.rb
+++ b/lib/stack_master/command.rb
@@ -27,6 +27,12 @@ module StackMaster
       end
     end
 
+    def initialize(config, stack_definition, options = Commander::Command::Options.new)
+      @config = config
+      @stack_definition = stack_definition
+      @options = options
+    end
+
     def success?
       @failed != true
     end
--- a/lib/stack_master/command.rb
+++ b/lib/stack_master/command.rb
@@ -36,6 +36,13 @@ module StackMaster
     def error_message(e)
       msg = "#{e.class} #{e.message}"
       msg << "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
+      if defined?(@options)
+        if @options&.trace
+          msg << "\n" << e.full_message
+        else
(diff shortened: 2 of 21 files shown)


lib/stack_master/template_compiler.rb
```diff
@@ -11,7 +11,7 @@ def self.compile(config, template_compiler, template_dir, template, compile_time
       compiler.require_dependencies
       compiler.compile(template_dir, template, compile_time_parameters, compiler_options)
     rescue StandardError => e
-      raise TemplateCompilationFailed.new("Failed to compile #{template} with error #{e}.\n#{e.backtrace}")
+      raise TemplateCompilationFailed, "Failed to compile #{template}"
```
orien: Remove the backtrace from the `TemplateCompilationFailed` error message. This was unreadable.

Fixes 260.

lib/stack_master/command.rb
```diff
@@ -36,9 +36,26 @@ def success?
     def error_message(e)
       msg = "#{e.class} #{e.message}"
       msg << "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
+      if defined?(@options)
+        if @options&.trace
+          msg << "\n#{backtrace(e)}"
+        else
+          msg << "\n Use --trace to view backtrace"
```
orien: Print the backtrace if the `--trace` option is provided. Otherwise include a message to inform the user to use `--trace` if more details are needed.

lib/stack_master/commands/lint.rb
```diff
@@ -9,6 +9,7 @@ class Lint
       def initialize(config, stack_definition, options = {})
         @config = config
         @stack_definition = stack_definition
+        @options = options
```
orien: Some commands weren't saving the options as an instance variable.

lib/stack_master/command.rb
```diff
@@ -27,6 +27,12 @@ def perform
       end
     end
 
+    def initialize(config, stack_definition, options = Commander::Command::Options.new)
+      @config = config
+      @stack_definition = stack_definition
+      @options = options
+    end
```
orien: Pull up `initialize` method into `Command` module. Most of the commands initialize with these params. There are a few that differ though:

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/commands/delete.rb#L7-L11

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/commands/init.rb#L8-L12

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/commands/list_stacks.rb#L10-L12

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/commands/status.rb#L10-L13

These are fine to override the default initializer.

orien:
> I wonder if we moved to having commands inherit from the `Command` class, instead of including it, we could avoid this meta-programming. It seems like they have the exact same `initialize` function, so we could dry that up as well as a nice side effect.

@patrobinson I like this suggestion. Unfortunately, upon investigation I found it would be quite a refactoring. Particularly, the way the `Command` module overrides the _subclass_ command object's `perform` method.

https://github.com/envato/stack_master/blob/d598e12de421ac7510e0e1bf1dd2e46d5bb88c91/lib/stack_master/command.rb#L21-L27

As a step along that path, I've pulled up the `initialize` method into the `Command` module to DRY up the concrete command classes.

lib/stack_master/cli.rb
```diff
@@ -71,7 +71,7 @@ def execute!
           unless args.size == 2
             say "Invalid arguments. stack_master init [region] [stack_name]"
           else
-            StackMaster::Commands::Init.perform(options.overwrite, *args)
+            StackMaster::Commands::Init.perform(options, *args)
```
orien: Ensure the options are passed into all commands.

lib/stack_master/commands/apply.rb
```diff
@@ -8,14 +8,10 @@ class Apply
...
-        @config = config
-        @s3_config = stack_definition.s3
-        @stack_definition = stack_definition
+      def initialize(*_args)
+        super
+        @s3_config = @stack_definition.s3
         @from_time = Time.now
-        @options = options
-        @options.on_failure ||= nil
-        @options.yes_param ||= nil
```
orien: I removed these two lines as they were a no-op: `Commander::Command::Options` instances will return `nil` for any unset CLI options.

lib/stack_master/cli.rb
```diff
@@ -178,7 +178,7 @@ def execute!
           options.defaults config: default_config_file
           say "Invalid arguments. stack_master tidy" and return unless args.size == 0
           config = load_config(options.config)
-          StackMaster::Commands::Tidy.perform(config)
+          StackMaster::Commands::Tidy.perform(config, options)
```
patrobinson: Shouldn't the signature be `config, nil, option` ?

lib/stack_master/commands/delete.rb
```diff
@@ -4,10 +4,11 @@ class Delete
       include Command
       include StackMaster::Prompter
 
-      def initialize(region, stack_name)
+      def initialize(region, stack_name, options)
         @region = region
         @stack_name = stack_name
         @from_time = Time.now
+        @options = options
```
patrobinson: Should this call `super`?

lib/stack_master/command.rb
```diff
@@ -36,9 +42,26 @@ def success?
     def error_message(e)
       msg = "#{e.class} #{e.message}"
       msg << "\n Caused by: #{e.cause.class} #{e.cause.message}" if e.cause
+      if defined?(@options)
```
patrobinson: Now should we define an accessor method
```
def options
  @options ||= {}
end
```

So we don't have to check if it's defined?

lib/stack_master/cli.rb
orien: You're right.

lib/stack_master/command.rb
orien: Added in 9621019.

lib/stack_master/cli.rb
orien: Fixed in 6256955

lib/stack_master/commands/delete.rb
orien: Changed in 239fa25.

Issue #262, opened 2018-12-18
Parameters are named with underscores, but stacks are named with dashes


mipearson:
Quoting the docs:

```

Parameters are loaded from multiple YAML files, merged from the following lookup paths from bottom to top:

parameters/[stack_name].yaml
parameters/[stack_name].yml
parameters/[region]/[underscored_stack_name].yaml
parameters/[region]/[underscored_stack_name].yml
parameters/[region_alias]/[underscored_stack_name].yaml
parameters/[region_alias]/[underscored_stack_name].yml
```

This is confusing when navigating your editor trying to find `foo-bar-baz` when you actually want `foo_bar_baz`, as in the main `stack_master.yml` it's defined with dashes.

I want to add parameter resolving that looks for dashes before underscores and make it the new default, perhaps deprecating underscore parameter files for StackMaster 2.

mbklaene:
Agree! I spent an hour trying to figure this out when I decided to rename my stack and assumed the parameter file would match the template.

PR #321, opened 2020-03-18
Dashes in parameter filenames

orien:
When a stack name includes a dash (`-`), the corresponding parameter filenames need this replaced with an underscore (`_`). This has produced much confusion (see #262).

I propose that when a stack name includes a dash, StackMaster should support a dash in the corresponding parameter filenames. For backwards compatibility, we should continue to support an underscore also.

In addition, the `stack_master init` command should produce parameter files that match the provided stack name.

### DIFF
--- a/lib/stack_master/commands/init.rb
+++ b/lib/stack_master/commands/init.rb
@@ -24,8 +24,8 @@ module StackMaster
       def check_files
         @stack_master_filename = "stack_master.yml"
         @stack_json_filename = "templates/#{@stack_name}.json"
-        @parameters_filename = File.join("parameters", "#{underscored_stack_name}.yml")
-        @region_parameters_filename = File.join("parameters", @region, "#{underscored_stack_name}.yml")
+        @parameters_filename = File.join("parameters", "#{@stack_name}.yml")
+        @region_parameters_filename = File.join("parameters", @region, "#{@stack_name}.yml")
 
         if !@options.overwrite
           [@stack_master_filename, @stack_json_filename, @parameters_filename, @region_parameters_filename].each do |filename|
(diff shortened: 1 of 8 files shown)


lib/stack_master/stack_definition.rb
```diff
@@ -105,20 +105,20 @@ def s3_configured?
...
 
     def default_parameter_glob
-      File.join(base_dir, 'parameters', "#{underscored_stack_name}.y*ml")
+      File.join(base_dir, 'parameters', "#{stack_name_glob}.y*ml")
     end
 
-    def underscored_stack_name
-      stack_name.gsub('-', '_')
+    def stack_name_glob
+      stack_name.gsub('-', '[-_]')
```
orien: Instead of substituting dash for underscore, substitute a glob so both dash and underscore will be used when searching the filesystem for corresponding parameter files.

lib/stack_master/commands/init.rb
```diff
@@ -24,8 +24,8 @@ def perform
       def check_files
         @stack_master_filename = "stack_master.yml"
         @stack_json_filename = "templates/#{@stack_name}.json"
-        @parameters_filename = File.join("parameters", "#{underscored_stack_name}.yml")
-        @region_parameters_filename = File.join("parameters", @region, "#{underscored_stack_name}.yml")
+        @parameters_filename = File.join("parameters", "#{@stack_name}.yml")
+        @region_parameters_filename = File.join("parameters", @region, "#{@stack_name}.yml")
```
orien: `stack_master init` uses the stack name as provided on the CLI when naming parameter files.

PR #324, opened 2020-03-20
Fix role assumer region issue

petervandoros:
#### Context

There is an error when assuming a role without having configured a default AWS region in your environment:

```
Executing apply on my-stack in us-east-1
Aws::Errors::MissingRegionError: missing region; use :region option or export region name to ENV['AWS_REGION']
```

#### Changes

- Use the CloudFormation driver's region, which includes fallbacks when it's not specified, when assuming a role

#### Considerations

Which region is used doesn't actually matter when assuming a role, as IAM and STS are global services.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,14 @@ The format is based on [Keep a Changelog], and this project adheres to
 - `stack_master apply` prints names of parameters with missing values
   ([#322]).
 
+### Fixed
+
+- Error assuming role when default aws region not configured in the
+  environment ([324])
+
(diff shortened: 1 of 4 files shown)


orien:
It seems the specs are suffering from a similar problem as described in the PR description: when a default region is defined in `~/.aws/config` the tests pass (a string value is passed to the `Aws::AssumeRoleCredentials` initialiser). But if no default region is defined, the tests fail (`nil` is passed to the `Aws::AssumeRoleCredentials`) as exhibited in CI.

PR #325, opened 2020-03-30
Add support for specifying account aliases in the allowed_accounts stack definition property

petervandoros:
Add support for specifying account aliases in the allowed_accounts stack definition property.


features/apply_with_allowed_accounts.feature
```diff
@@ -53,3 +56,22 @@ Feature: Apply command with allowed accounts
...
     Then the exit status should be 0
+
+  Scenario: Run apply with stack specifying allowed account alias
+    Given I stub the following stack events:
+      | stack_id | event_id | stack_name  | logical_resource_id | resource_status | resource_type              | timestamp           |
+      | 1        | 1        | myapp-cache | myapp-cache         | CREATE_COMPLETE | AWS::CloudFormation::Stack | 2020-10-29 00:00:00 |
+    When I use the account "44444444" with alias "my-account-alias"
+    And I run `stack_master apply us-east-1 myapp-cache`
+    And the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} myapp-cache AWS::CloudFormation::Stack CREATE_COMPLETE/
+    Then the exit status should be 0
```
orien: We should use `Then` for lines with assertions:

```suggestion
    Then the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} myapp-cache AWS::CloudFormation::Stack CREATE_COMPLETE/
    And the exit status should be 0
```

lib/stack_master/identity.rb
```diff
@@ -1,16 +1,21 @@
 module StackMaster
   class Identity
-    def running_in_allowed_account?(allowed_accounts)
-      allowed_accounts.nil? || allowed_accounts.empty? || allowed_accounts.include?(account)
+    def running_in_account?(accounts)
```
orien: ❤️

lib/stack_master/identity.rb
```diff
@@ -1,16 +1,21 @@
...
+        contains_account_alias?(accounts)
     end
 
     def account
       @account ||= sts.get_caller_identity.account
     end
 
-    private
+    def account_aliases
+      @aliases ||= iam.list_account_aliases.account_aliases
```
orien: With the addition of this new API call, I'm assuming StackMaster requires additional permissions. Perhaps we should include this in the changelog.

features/apply_with_allowed_accounts.feature
```diff
@@ -53,3 +56,22 @@ Feature: Apply command with allowed accounts
...
+
+  Scenario: Run apply with stack specifying disallowed account alias
+    Given I stub the following stack events:
+      | stack_id | event_id | stack_name | logical_resource_id | resource_status | resource_type              | timestamp           |
+      | 1        | 1        | myapp-db   | myapp-db            | CREATE_COMPLETE | AWS::CloudFormation::Stack | 2020-10-29 00:00:00 |
+    When I use the account "11111111" with alias "an-account-alias"
+    And I run `stack_master apply us-east-1 myapp-db`
```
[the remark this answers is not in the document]
petervandoros: Fixed in bb773ab.

features/apply_with_allowed_accounts.feature
petervandoros: Fixed in 6061e54.

lib/stack_master/identity.rb
petervandoros: Added in 1fc44b7.

lib/stack_master/identity.rb
```diff
@@ -19,5 +24,17 @@ def region
...
+    def iam
+      @iam ||= Aws::IAM::Client.new(region: region)
+    end
+
+    def contains_account_id?(ids)
+      ids.include?(account)
+    end
+
+    def contains_account_alias?(aliases)
+      account_aliases.any? { |account_alias| aliases.include?(account_alias) }
```
orien, line 37: Perhaps this should only be checked (and the API call made) if the provided values look like aliases, ie. they're not numbers.

petervandoros: I thought about this too, but the problem with this approach is that it only covers the case where the check is performed, not when the error message is being generated in the cli.rb file.

I think the way to cover all cases is to rescue the permissions error in the `account_aliases` method and hard fail with a useful error message.

orien: 👍 Let's do that.

petervandoros:
@patrobinson @orien I added a useful error message when the `iam:ListAccountAliases` permission is missing in 8a528b7.

Please have a look 😄 .

Issue #238, opened 2018-06-03
StackMaster 2.0 Proposal: Environments


patrobinson:
Hello one and all,

In October StackMaster will turn three and we've decided the next major goal should be a 2.0 release. This is an opportunity to put all options on the table, including those that break the API. This issue is a place to discuss the proposal to solve the hottest topic in StackMaster today, Environments and their current implementation, Region Aliases. Feel free to comment here on this proposal as well as potential alternatives.

## Problem Statement
It’s common to have other environments in addition to Production, which replicate all or some of Production stacks. We can share Parameters between environments, but it may also be beneficial to share Parameters within an environment (i.e. log server address).

The current method of using region aliases is limited. Users cannot create multiple environments in the same region and conflating the concept of an environment and region can make it difficult for multi-region services.

For the past year or more I've encouraged we defer any decision on this problem until we understood it better. Having found this problem in many seperate implementations across the Envato business I feel we have enough experience with the problem to make a well informed decision, but I'd love to hear from others who have also felt this pain including @johnf @rbayerl @nitehawk @berniedurfee-ge @flyinbutrs .

Relevant Issues and PRs:
#109
#180
#221 
#115 

## Current Solutions

As discussed we can use region-aliases to deploy environments to different regions.

Alternatively you can have a seperate StackMaster path, with a different stack_master.yml file and parameter files, while linking to a shared template path. This makes it possible to have a directory structure like so:

```
- production/
-- stack_master.yml
-- parameters/
- staging/
-- stack_master.yml
-- parameters/
- templates/
```

However it introduces duplication between the stack_master.yml files.

## Proposed Solution

Instead of using a region to delineate an environment, I propose we use an AWS Account. While this is again prescriptive, it’s at least best practice as recommended by Amazon. We can either fetch the account number (list-account-aliases) or take an environment variable to tell us what account we are in. Each “environment” would have a directory (similar to that described above) and it’s own stack_master.yml file, which contained stack_defaults such as tags and stacks that are specific to that environment. There would also be a "master" stack_master.yml file defining stacks that are applicable to both environments, which would be merged with the environment configuration at run time.

```
- production/
-- stack_master.yml
-- parameters/
- stack_master.yml
- staging/
-- stack_master.yml
-- parameters/
- templates/
```

This would not break existing functionality, users could continue to use a single StackMaster config file for all environments or one for each environment. Only when the `environments:` key is encountered in the top level config file would this functionality take effect.

orien:
[...]

PR #330, opened 2020-04-20
Spike/WIP: Stack definition yaml files. Explicit parameter_files & parameters in stack definitions.

stevehodgkiss:
Some experimental changes to simplify StackMaster usage.

#### Stack definitions move from stack_master.yml to their own stack-name.yml

Stacks can be defined in their own YAML file and applied with `stack_master apply stack-name.yml`. `region` must be defined in `stack-name.yml` or `stack_master.yml`'s `stack_defaults`.

This adds flexibility in terms of how stacks are organised on the filesystem, and prevents `stack_master.yml` bloat.

#### Explicitly define parameter files locations with `parameter_files`

Adds the ability to explicitly define parameter file locations in stack definitions `parameter_files: [my-parameters.yml]`. The benefit here is less magic about where they're sourced from, and the ability to share parameter files in any way you see fit. For example, if we have `myapp-blue` and `myapp-green`, they could share a parameter file, whereas currently that's not possible without duplicating the parameters.

#### Add the ability to define parameters in stack definitions

Used with stacks in `stack_master.yml`, this would make the file extremely bloated. When stacks are in their own yaml file though, it makes them easier to read and understand in context rather than being split over multiple files.

These changes may remove the need to have any concept of environment/region alias as discussed as part of [StackMaster 2.0](https://github.com/envato/stack_master/issues/238).

#### Example

Here's a made up example using all these features:

```
cat myapp-web-blue.yml
template: myapp.rb
tags:
  Name: myapp-blue
parameter_files:
  - shared_parameters/myapp-web.yml
  - shared_parameters/myapp-web-production.yml
parameters:
  name: myapp-blue
  bucket_name:
    stack_output: myapp-blue/bucket_name

cat shared-parameters/myapp-web.yml
some_param: 'value'

cat shared-parameters/myapp-web-production.yml
some_other_param: 'value'

stack_master apply myapp-web-blue.yml
```

Thoughts?

### DIFF
--- a/features/apply_with_explicit_parameter_files.feature
+++ b/features/apply_with_explicit_parameter_files.feature
@@ -1,0 +1,54 @@
+Feature: Apply command with explicit parameter files
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stack_defaults:
+        tags:
+          Application: myapp
+        region: us-east-1
+      """
+    And a file named "myapp-web.yml" with:
+      """
+      template: myapp.rb
+      parameter_files:
+      - myapp-web-parameters.yml
+      """
+    And a file named "myapp-web-parameters.yml" with:
+      """
+      KeyName: my-key
+      """
+    And a directory named "templates"
+    And a file named "templates/myapp.rb" with:
+      """
+      SparkleFormation.new(:myapp) do
+        description "Test template"
+
+        parameters.key_name do
+          description 'Key name'
+          type 'String'
+        end
+
+        resources.instance do
+          type 'AWS::EC2::Instance'
+          properties do
+            image_id 'ami-0080e4c5bc078760e'
+            instance_type 't2.micro'
+          end
+        end
+      end
+      """
+
+  Scenario: Run apply with stack-name.yml and explicit parameter files
+    Given I stub the following stack events:
+      | stack_id | event_id | stack_name | logical_resource_id | resource_status | resource_type              | timestamp           |
+      | 1        | 1        | myapp-web  | myapp-web           | CREATE_COMPLETE | AWS::CloudFormation::Stack | 2020-10-29 00:00:00 |
+    When I run `stack_master apply myapp-web.yml --trace`
+    Then the output should match /2020-10-29 00:00:00 (\+|\-)[0-9]{4} myapp-web AWS::CloudFormation::Stack CREATE_COMPLETE/
+    And the output should contain all of these lines:
+      | Stack diff:          |
+      | +    "Instance": {   |
+      | Parameters diff:     |
+      | KeyName: my-key      |
+      | Proposed change set: |
+    And the exit status should be 0
--- a/features/apply_with_stack_definition_parameters.feature
+++ b/features/apply_with_stack_definition_parameters.feature
@@ -1,0 +1,46 @@
+Feature: Apply command with stack definition parameters
+
+  Background:
+    Given a file named "stack_master.yml" with:
+      """
+      stacks:
+        us-east-1:
+          myapp_web:
+            template: myapp.rb
+            parameters:
+              KeyName: my-key
+      """
+    And a directory named "templates"
+    And a file named "templates/myapp.rb" with:
+      """
+      SparkleFormation.new(:myapp) do
+        description "Test template"
+
+        parameters.key_name do
+          description 'Key name'
+          type 'String'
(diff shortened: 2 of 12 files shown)


joesustaric:
Interesting @stevehodgkiss .. so just clarifying some observations.

In the example, the `myapp-web-blue.yml` file specifies 2 parameter files. I'm assuming that behaviour is the same as current. eg they inherit and will subsequently override any params.

There are production params being inherited in the example (determined via `- shared_parameters/myapp-web-production.yml`), so I'm assuming this example is a prod stack which would then need to be organised in some logical file system grouping . i.e `myapp-web-blue/production` 

If thats true then there would be something like `myapp-web-blue/staging` . Would that mean the stack definition for `myapp-web-blue.yml` would need to exist in both folders?

stevehodgkiss:
> In the example, the myapp-web-blue.yml file specifies 2 parameter files. I'm assuming that behaviour is the same as current. eg they inherit and will subsequently override any params.

Yep, they'd be merged from first to last.

> There are production params being inherited in the example (determined via - shared_parameters/myapp-web-production.yml), so I'm assuming this example is a prod stack which would then need to be organised in some logical file system grouping . i.e myapp-web-blue/production

Yep, that's the idea. You can structure the stack yaml files on the filesystem however you see fit - for example, separate directories for staging/production. And you can share parameter files in any way that makes sense. More flexibility, less magic.

> If thats true then there would be something like myapp-web-blue/staging . Would that mean the stack definition for myapp-web-blue.yml would need to exist in both folders?

Yep, in the same way that there'd be 2 definitions in a `stack_master.yml` file with the current approach.

The PR atm supports the existing approach and this one.

joesustaric:
Yep great gotcha, thanks for clarifying.

Seems like a good idea. 
I've certainly seen some people new to `stack_master` have some confusion when navigating the templates, tracing back parameters and navigating environments from alias' etc..
So a big +1 to removing some of the magic.

patrobinson:
this is going in the right direction

I like the idea that environments/region_defaults can be implemented without a massive refactor like i was thinking. I agree that it's better to be explicit about which parameter files are loaded.

However i dislike allowing both approaches and merging the results from the existing hierarchical lookup with explicit parameter files as well as parameters defined in a stack definiition file. This is adding arguably more magic as you can now have an unlimited number of files that can define parameters.

I also am not sure if adding another file, the stack definition file, to the mix helps, as now you can have 3 files define the behaviour of a stack instead of just 2. This implies some more magic, while removing some other.

I'd really like to nail down a name for and scope for a concept like what @grassdog referred to in #256 . This could be defined in a single file, like you've implemented. Then some "targets" for that stack, with the stack name and region as well as override parameters.

The more I think about it, the more getting rid of a single stack_master.yml file with a directory of parameter files makes more sense.

Imagine your scenario:

myapp-web.yml:
```yaml
template: myapp.rb
parameter_files:
  - shared_parameters/myapp-web.yml
targets:
  blue:
    region: us-east-1
    allowed_accounts: '987654321'
    tags:
      Name: myapp-blue
    parameter_files:
      - shared_parameters/myapp-web-production.yml
  green:
    region: us-east-1
    allowed_accounts: '987654321'
    tags:
      Name: myapp-blue
    parameter_files:
      - shared_parameters/myapp-web-production.yml
  staging:
    region: us-east-1
    allowed_accounts: '123456789'
    tags:
      Name: myapp-blue
  parameters:
    staging: 'true'
    bucket_name:
      stack_output: foo/bucket_name
```

```bash
stack_master apply myapp-web blue
```

But this still involves the giant refactor I've been procrastinating over for 18 months

patrobinson:
> Yeah, I was thinking about removing support for defining stacks in stack_master.yml and the magic parameter lookups. I haven't used this approach enough to know for sure that's the direction I'd like, but I think it is.

Should we enable it behind a feature flag? Have the ability to turn on/off explicit parameters, defaulting to off.

With "stack definition" files my loosely held opinion is I'd rather have them be standalone, i.e. if a stack definition file is specified the stack_master.yml file is not referred to at all.

> What benefit would modelling targets inside StackMaster bring vs allowing the user to lay it out like that manually?

Nothing tangible at present, but I believe in the future when we can define multiple stacks a "target" applies then it'll prove more useful. For the purposes of this discussion it's not particularly useful as you have more concrete problems you want to solve.

PR #323, opened 2020-03-20
Validate command checks for missing parameter values

orien:
An implementation of the improvement suggested in #189.

There is some custom code in the `apply` command that checks for parameters with no values. If blank parameters are found, it aborts the process and prints a friendly message.

In this proposal, the code has been extracted to a separate class, `ParameterValidator` and the `validate` command modified to use this check. It prints out the same helpful message if it finds this issue.

```
myapp_vpc: invalid
Empty/blank parameters detected. Please provide values for these parameters:
 - ParamOne
 - ParamTwo
Parameters will be read from files matching the following globs:
 - parameters/myapp_vpc.y*ml
 - parameters/us-east-1/myapp_vpc.y*ml
```

PR #331, opened 2020-05-01
Add `stack_master validate --no-validate-template-parameters` option

orien:
Since version 2.4.0 and #323, the `stack_master validate` validates the presence of template parameter values. This new feature is valuable, has the potential to cause pain in the upgrade process as it:

- attempts to resolve parameters
- requires extra IAM permissions, depending on the parameter resolving features used by the given stack

To ease the upgrade path, I propose an option to disable this new functionality.

```
stack_master validate --no-validate-template-parameters <region> <stack>
```

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -46,7 +46,7 @@ module StackMaster
         c.option '--on-failure ACTION', String, "Action to take on CREATE_FAILURE. Valid Values: [ DO_NOTHING | ROLLBACK | DELETE ]. Default: ROLLBACK\nNote: You cannot use this option with Serverless Application Model (SAM) templates."
         c.option '--yes-param PARAM_NAME', String, "Auto-approve stack updates when only parameter PARAM_NAME changes"
         c.action do |args, options|
-          options.defaults config: default_config_file
+          options.default config: default_config_file
           execute_stacks_command(StackMaster::Commands::Apply, args, options)
         end
       end
@@ -56,7 +56,7 @@ module StackMaster
         c.summary = 'Displays outputs for a stack'
         c.description = "Displays outputs for a stack"
         c.action do |args, options|
-          options.defaults config: default_config_file
+          options.default config: default_config_file
           execute_stacks_command(StackMaster::Commands::Outputs, args, options)
         end
       end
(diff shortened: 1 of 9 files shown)


lib/stack_master/cli.rb
```diff
@@ -46,7 +46,7 @@ def execute!
         c.option '--on-failure ACTION', String, "Action to take on CREATE_FAILURE. Valid Values: [ DO_NOTHING | ROLLBACK | DELETE ]. Default: ROLLBACK\nNote: You cannot use this option with Serverless Application Model (SAM) templates."
         c.option '--yes-param PARAM_NAME', String, "Auto-approve stack updates when only parameter PARAM_NAME changes"
         c.action do |args, options|
-          options.defaults config: default_config_file
+          options.default config: default_config_file
```
orien, line 49: The default option values are specified via the `default` method.

lib/stack_master/cli.rb
```diff
@@ -128,8 +128,9 @@ def execute!
         c.summary = 'Validate a template'
         c.description = 'Validate a template'
         c.example 'validate a stack named myapp-vpc in us-east-1', 'stack_master validate us-east-1 myapp-vpc'
+        c.option '--[no-]validate-template-parameters', 'Validate template parameters. Default: validate'
         c.action do |args, options|
-          options.defaults config: default_config_file
+          options.default config: default_config_file, validate_template_parameters: true
```
orien, line 133: With no option specified, resolve and validate the template parameters.

lib/stack_master/stack.rb
```diff
@@ -75,6 +75,25 @@ def self.generate(stack_definition, config)
           stack_policy_body: stack_policy_body)
     end
 
+    def self.generate_without_parameters(stack_definition, config)
```
orien, line 78: A new way to generate a `Stack` instance without resolving template parameters.

petervandoros, line 78: Given compile time parameters are resolved, consider renaming the method to be slightly more accurate. Perhaps something like `generate_without_template_parameters`?

orien: I'm in two minds about this. The `Stack` instance this generates has a `parameters` property which only includes the template parameters. Compile-time parameters don't seem to be a thing once the `Stack` instance has been generated.

petervandoros: > I'm in two minds about this. The Stack instance this generates has a parameters property which only includes the template parameters.

Happy to leave it as is. I don't feel strongly about it 😄 .

> Compile-time parameters don't seem to be a thing once the Stack instance has been generated.

Correct. They're used to _generate_ the template. Template parameters are _part of_ the template and evaluated at apply time by the CloudFormation service.

orien: 👍 

I think I'll leave it as is. Differentiating between the two types of parameters here seems to be exposing internal implementation of the method. The output `Stack` instance names the template parameters, 'parameters', so I'll follow that convention.

PR #335, opened 2020-05-27
Fix json template detection with extra leading whitespace

liamdawson:
When running template format detection on a template body, extra leading whitespace (e.g. `"\n {"`) would lead to the template being detected as YAML.

This in turn resulted in guaranteed diff mismatches if the last applied template began with `"\n {"` and wasn't prettily formatted.

Also added specs for `StackMaster::TemplateUtils.identify_template_format` to test current behaviour and identify regressions.

### DIFF
--- a/lib/stack_master/template_utils.rb
+++ b/lib/stack_master/template_utils.rb
@@ -2,14 +2,14 @@ module StackMaster
   module TemplateUtils
     MAX_TEMPLATE_SIZE = 51200
     MAX_S3_TEMPLATE_SIZE = 460800
+    # Matches if the first non-whitespace character is a '{', handling cases
+    # with leading whitespace and extra (whitespace-only) lines.
+    JSON_IDENTIFICATION_PATTERN = Regexp.new('\A\s*{', Regexp::MULTILINE)
 
     extend self
 
     def identify_template_format(template_body)
-      # if the first non-whitespace character across all lines is a '{'
-      json_pattern = Regexp.new('\A\s*{', Regexp::MULTILINE)
-
-      if template_body =~ json_pattern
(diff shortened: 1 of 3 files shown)


lib/stack_master/template_utils.rb
```diff
@@ -6,8 +6,14 @@ module TemplateUtils
     extend self
 
     def identify_template_format(template_body)
-      return :json if template_body =~ /^{/x # ignore leading whitespaces
-      :yaml
+      # if the first non-whitespace character across all lines is a '{'
+      json_pattern = Regexp.new('\A\s*{', Regexp::MULTILINE)
```
stevehodgkiss, line 10: Could this be stored in a constant?

liamdawson: I reflexively went to do that, but wasn't sure if we want to take the performance hit to compile it whenever we include the file. Doubt it's honestly a performance cost worth worrying about, though?

stevehodgkiss: Yeah I doubt it would make a difference in the typical use case - `stack_master apply us-east-1 my-stack`.

orien:
Could you please add an entry in the changelog?

lib/stack_master/template_utils.rb
liamdawson: https://github.com/envato/stack_master/pull/335/commits/ec99ebdb2ad9f578f860a30fdcb55dcb91ab8e1c

liamdawson:
@orien Added in https://github.com/envato/stack_master/pull/335/commits/70f3f3ce231e9e6b9733b1f41844a54cdeb5ab05

orien:
🙇🏻‍♂️ Great! Thank you.

PR #336, opened 2020-06-01
Add a conditional replace explanation

viraptor:
Apart from well defined changes, resources may be also conditionally replaced.
This should be annotated in the proposed changes.
Also change the colour to red for potential replacements.

### DIFF
--- a/lib/stack_master/change_set.rb
+++ b/lib/stack_master/change_set.rb
@@ -68,8 +68,11 @@ io.puts "========================================"
     private
 
     def display_resource_change(io, resource_change)
-      action_name = if resource_change.replacement == 'True'
+      action_name = case resource_change.replacement
+                    when 'True'
                       'Replace'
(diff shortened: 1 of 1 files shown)


viraptor:
Warnings: No tests. I'm not sure how to test this and I sank enough time flailing around. Ideas welcome.

jacobbednarz:
What about reusing the AWS terms here? Using `Conditionally replace` instead of `Maybe replace`?

lib/stack_master/change_set.rb
```diff
@@ -101,7 +106,7 @@ def action_color(action_name)
         :green
       when 'Modify'
         :yellow
-      when 'Remove', 'Replace'
+      when 'Remove', 'Replace', 'Maybe replace'
```
stevehodgkiss, line 109: Does this need to be `Conditional replace`?

PR #338, opened 2020-06-10
Add parameter_files and parameters configuration options to stack definition

stevehodgkiss:
Following on from an earlier spike, but without the stack definition moving to its own file for better backwards compatibility. https://github.com/envato/stack_master/pull/330.

#### Added `parameters_dir` config

This matches what we have for `template_dir` and is useful for the `parameter_files` feature.

#### `parameter_files` - an array of explicit parameter file locations based from `parameter_dir`

Adds the ability to explicitly define parameter file locations in stack definitions. It's more obvious this way, and it can also be used to massively reduce duplication between stacks that follow a blue/green setup. For example, if we have `myapp-blue` and `myapp-green`, they could share a parameter file for the majority of parameters, whereas currently that's not possible without duplicating all the parameters.

```yaml
stacks:
  us-east-1:
    myapp-blue:
      parameter_files:
        - myapp.yml
        - myapp-blue.yml
      template: myapp.rb
```

#### `parameters` hash key can add parameters inline with the rest of the stack definition

This can be useful when switching to a new stack that only needs a small change in parameters:

```yaml
stacks:
  us-east-1:
    aws-elasticsearch-green:
      parameter_files:
        - aws-elasticsearch.yml
      template: aws-elasticsearch.rb
    aws-elasticsearch-blue:
      parameter_files:
        - aws-elasticsearch.yml
      parameters:
        engine_version: 7.4
      template: aws-elasticsearch.rb
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,12 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ### Added
 
-- `parameter_dir` is now configurable to match the existing `template_dir`.
-- A `parameter_files` array allows configuring an array of *explicit* parameter
-  files relative to `parameter_dir`. If this option is specified, automatic
-  parameter files based on region and the stack name will not be used.
-- A `parameters` hash key allows defining parameters directly on the stack
-  definition rather than requiring an external parameter file.
+- `parameters_dir` is now configurable to match the existing `template_dir`.
+- `parameter_files` configures an array of parameter files relative to
+  `parameters_dir` that will be used instead of automatic parameter file globs
+  based on region and stack name.
+- `parameters` configures stack parameters directly on the stack definition
+  rather than requiring an external parameter file.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog], and this project adheres to
 [Keep a Changelog]: https://keepachangelog.com/en/1.0.0/
 [Semantic Versioning]: https://semver.org/spec/v2.0.0.html
 
-## [Unreleased]
+## [2.7.0] - 2020-06-15
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- `parameter_dir` is now configurable to match the existing `template_dir`.
+- A `parameter_files` array allows configuring an array of *explicit* parameter
+  files relative to `parameter_dir`. If this option is specified, automatic
+  parameter files based on region and the stack name will be be used.
+- A `parameters` hash key allows defining parameters directly on the stack
+  definition rather than requiring an external parameter file.
+
 ### Fixed
 
 - JSON template bodies with whitespace on leading lines would incorrectly be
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: lib/stack_master/stack_definition.rb [hunk 1]
```diff
@@ -34,6 +35,7 @@ def initialize(attributes = {})
       @additional_parameter_lookup_dirs ||= []
       @template_dir ||= File.join(@base_dir, 'templates')
       @allowed_accounts = Array(@allowed_accounts)
+      @parameters ||= {}
```
petervandoros, line 42: Why use the conditional assignment operator in the initializer? Is `@parameters` expected to be set before the initializer is run?

CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- `parameter_dir` is now configurable to match the existing `template_dir`.
+- A `parameter_files` array allows configuring an array of *explicit* parameter
+  files relative to `parameter_dir`. If this option is specified, automatic
+  parameter files based on region and the stack name will be be used.
```
petervandoros, line 18: Typo?
```suggestion
  parameter files based on region and the stack name will not be used.
```

features/step_definitions/stack_steps.rb
```diff
@@ -73,3 +73,9 @@ def extract_hash_from_kv_string(string)
   file = StackMaster.s3_driver.find_file(bucket: bucket, object_key: key)
   expect(file).to eq body
 end
+
+When(/^an S3 file in bucket "([^"]*)" with key "([^"]*)" exists with JSON content:$/) do |bucket, key, body|
+  file = StackMaster.s3_driver.find_file(bucket: bucket, object_key: key)
+  parsed_file = JSON.parse(file)
+  expect(parsed_file).to eq JSON.parse(body)
+end
```
patrobinson, line 81: Not sure what this test achieves?

lib/stack_master/stack_definition.rb
stevehodgkiss: The call to `super` above sets values from the yaml file, so it would either be set from that point or not, in which case the default is `{}`.

features/step_definitions/stack_steps.rb
stevehodgkiss: It checks the contents match by comparing parsed hashes rather than strings. For some reason apply_with_s3.feature started failing, and master and has the same issue - https://travis-ci.org/github/envato/stack_master/jobs/692383878#L2514 Maybe it's related to a sparkleformation upgrade?

lib/stack_master/stack_definition.rb
```diff
@@ -101,20 +112,26 @@ def s3_configured?
       !s3.nil?
     end
 
+    def parameter_files
+      Array(@parameter_files).map do |file|
+        File.expand_path(File.join(parameters_dir, file))
```
orien, line 117: I wonder if this can be simplified to the following:

```ruby
File.expand_path(file, parameters_dir)
```

stevehodgkiss:
> Do you think we should also support compile-time parameters in the stack_master.yml?

I think they would be supported already due to the way `compile_time_parameters` are implemented (as a special parameter).

```
parameters:
  param: 1
  compile_time_parameters:
    param: 1
```

orien:
Oh I see. I tried it and you're quite right, it does just work!

PR #339, opened 2020-06-18
Add drift command to display resources that have been changed outside of CFN

stevehodgkiss:
Uses the CFN drift detection API to trigger drift detection & display drifted resources:

stackmaster drift feature

### DIFF
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -56,6 +56,7 @@ module StackMaster
   module Commands
     autoload :TerminalHelper, 'stack_master/commands/terminal_helper'
     autoload :Apply, 'stack_master/commands/apply'
+    autoload :Drift, 'stack_master/commands/drift'
     autoload :Events, 'stack_master/commands/events'
     autoload :Outputs, 'stack_master/commands/outputs'
     autoload :Init, 'stack_master/commands/init'
@@ -197,5 +198,29 @@ module StackMaster
 
   def stderr=(io)
     @stderr = io
+  end
+
+  def colorize(text, color)
+    if colorize?
+      Rainbow(text).color(color)
+    else
+      text
+    end
+  end
+
+  def colorize?
+    ENV.fetch('COLORIZE') { 'true' } == 'true'
+  end
+
+  def display_colorized_diff(diff)
+    diff.each_line do |line|
+      if line.start_with?('+')
+        stdout.print colorize(line, :green)
+      elsif line.start_with?('-')
+        stdout.print colorize(line, :red)
+      else
+        stdout.print line
+      end
+    end
   end
 end
--- a/lib/stack_master.rb
+++ b/lib/stack_master.rb
@@ -24,6 +24,7 @@ module StackMaster
   autoload :CLI, 'stack_master/cli'
   autoload :CtrlC, 'stack_master/ctrl_c'
   autoload :Command, 'stack_master/command'
+  autoload :Diff, 'stack_master/diff'
   autoload :VERSION, 'stack_master/version'
   autoload :Stack, 'stack_master/stack'
   autoload :Prompter, 'stack_master/prompter'
@@ -210,17 +211,5 @@ module StackMaster
 
   def colorize?
     ENV.fetch('COLORIZE') { 'true' } == 'true'
-  end
-
-  def display_colorized_diff(diff)
-    diff.each_line do |line|
-      if line.start_with?('+')
-        stdout.print colorize(line, :green)
-      elsif line.start_with?('-')
-        stdout.print colorize(line, :red)
-      else
-        stdout.print line
-      end
-    end
   end
 end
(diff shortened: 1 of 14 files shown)


lib/stack_master/cli.rb
```diff
@@ -215,6 +215,17 @@ def execute!
         end
       end
 
+      command :drift do |c|
+        c.syntax = 'stack_master drift [region_or_alias] [stack_name]'
+        c.summary = 'Detects and displays stack drift'
```
jacobbednarz, line 220: Is it assumed here that "displays stack drift" means per Cloudformation standards and not what stack master has?

joesustaric: Yup I have similar question, I have not looked at the code but I assume this is a drift between the current aws stack definition vs the actual stack resources and config. (i.e not the stack master code)

P.S - big +1 to including this stuff 🏆

jacobbednarz: I assumed as much too but do wonder whether other users of this are on the same page.

petervandoros: How about tweaking the summary to be more explicit?
```suggestion
        c.summary = 'Detects and displays stack drift using the CloudFormation Drift API'
```

patrobinson: Yep I think we need documentation to clarify to users clearly what the difference between `diff` and `drift` is, as it's easily confused.

lib/stack_master/commands/drift.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      private
+
+      def cf
+        @cf ||= StackMaster.cloud_formation_driver
+      end
+
+      def display_drift(drift)
+        color = drift_color(drift)
+        puts colorize("#{drift.stack_resource_drift_status} #{drift.resource_type} #{drift.logical_resource_id} #{drift.physical_resource_id}", color)
```
patrobinson, line 37: I rarely care about line length but it's making it a little hard to review this PR.

lib/stack_master/commands/drift.rb
```diff
@@ -0,0 +1,115 @@
...
+      end
+
+      def stack_drift_status_color(stack_drift_status)
+        case stack_drift_status
+        when 'IN_SYNC'
+          :green
+        when 'DRIFTED'
+          :yellow
+        else
+          :blue
```
patrobinson, line 70: What other states are there?

lib/stack_master.rb
```diff
@@ -198,4 +199,28 @@ def stderr
...
+    diff.each_line do |line|
+      if line.start_with?('+')
+        stdout.print colorize(line, :green)
+      elsif line.start_with?('-')
+        stdout.print colorize(line, :red)
+      else
+        stdout.print line
+      end
+    end
+  end
```
patrobinson, line 225: It feels like this deserves it's own class

lib/stack_master/commands/drift.rb
stevehodgkiss: > String, one of "DRIFTED", "IN_SYNC", "UNKNOWN", "NOT_CHECKED"

stevehodgkiss:
> I wonder if we should fail if any of the states are not as expected? I think I prefer to just bail if it's not IN_SYNC/DRIFTED.

If there are no resource drifts to display in the second part (`cf.describe_stack_resource_drifts`), it will exit without displaying anything. I'm not sure how you end up in the `UNKNOWN` state, but there might information in the resource drifts to display, so I think it's reasonable to still attempt it 🤷 

> Other than that I would love to see some good practical examples of what the difference between drift and diff is in the README.

I've added some info about the 2 commands to the readme. I might update the readme with example screenshots after this has been merged and released. The apply demo is from before changesets were implemented, so it could do with an update.

lib/stack_master/commands/drift.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      def display_resource_drift(drift)
+        diff = ::StackMaster::Diff.new(before: prettify_json(drift.expected_properties),
+                        after: prettify_json(drift.actual_properties))
+        diff.display_colorized_diff
+      end
+
+      def prettify_json(string)
+        JSON.pretty_generate(JSON.parse(string)) + "\n"
+      rescue => e
```
orien, line 61: Nitpick: let's rescue a specific error:

```suggestion
      rescue => StandardError e
```

stevehodgkiss:
> Do you want to slap the frozen string literals comment on new files? (it's going to be enforced in ruby 3, so it will be needed anyway)

Not as part of this PR. I'd rather do all files in one sweep in a separate PR.

stevehodgkiss:
> Do you think it's worth a writing a feature test?

I've added an aruba integration test using rspec instead of cucumber. It uses the built in aws client library stubbing mechanism rather than the custom driver class used in cucumber. I think that style could be easier to read and maintain going forward.

PR #342, opened 2020-07-02
Add cfn_nag integration

simpsora:
[cfn_nag](https://github.com/stelligent/cfn_nag) is an open-source SAST (static application security testing) tool that checks CloudFormation templates for common patterns that may indicate a security problem.

While it is a Ruby library, it doesn't lend itself to direct integration.  For this reason the gem is installed from the gemspec but stack_master shells out to call the CLI command.

This PR adds a new `nag` command that runs `cfn_nag` on a given stack and outputs a report.

### DIFF
--- a/lib/stack_master/commands/nag.rb
+++ b/lib/stack_master/commands/nag.rb
@@ -1,0 +1,44 @@
+module StackMaster
+  module Commands
+    class Nag
+      include Command
+      include Commander::UI
+
+      def perform
+        unless cfn_nag_available
+          failed! 'Failed to run cfn_nag. You may need to install it using'\
+                  '`gem install cfn_nag`, or add it to $PATH.'\
+                  "\n"\
+                  '(See https://github.com/stelligent/cfn_nag'\
+                  ' for package information)'
+        end
+
+        if proposed_stack.template_format == :yaml
+          failed! 'cfn_nag doesn\'t support yaml formatted templates.'
+        end
+
+        Tempfile.open(['stack', "___#{stack_definition.stack_name}.#{proposed_stack.template_format}"]) do |f|
+          f.write(proposed_stack.template_body)
+          f.flush
+          system('cfn_nag', f.path)
+          puts "cfn_nag run complete"
+        end
+      end
+
+      private
+
+      def stack_definition
+        @stack_definition ||= @config.find_stack(@region, @stack_name)
+      end
+
+      def proposed_stack
+        @proposed_stack ||= Stack.generate(stack_definition, @config)
+      end
+
+      def cfn_nag_available
+        system('type -a cfn_nag >/dev/null 2>&1')
+        $?.exitstatus == 0
+      end
+    end
+  end
+end
--- a/lib/stack_master/commands/nag.rb
+++ b/lib/stack_master/commands/nag.rb
@@ -5,12 +5,14 @@ module StackMaster
       include Commander::UI
 
       def perform
-        Tempfile.open(['stack', "___#{stack_definition.stack_name}.#{proposed_stack.template_format}"]) do |f|
+        rv = Tempfile.open(['stack', "___#{stack_definition.stack_name}.#{proposed_stack.template_format}"]) do |f|
           f.write(proposed_stack.template_body)
           f.flush
           system('cfn_nag', f.path)
-          STDERR.puts "cfn_nag run complete"
+          $?.exitstatus
         end
+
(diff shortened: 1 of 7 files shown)


lib/stack_master/commands/nag.rb
```diff
@@ -0,0 +1,32 @@
+module StackMaster
+  module Commands
+    class Nag
+      include Command
+      include Commander::UI
+
+      def perform
+        if proposed_stack.template_format == :yaml
+          failed! 'cfn_nag doesn\'t support yaml formatted templates.'
```
viraptor, line 9: Can we do a temporary file with a yaml -> json conversion and run the test on it?

spec/stack_master/commands/nag_spec.rb
```diff
@@ -0,0 +1,42 @@
+RSpec.describe StackMaster::Commands::Nag do
+  let(:region) { 'us-east-1' }
+  let(:stack_name) { 'myapp-vpc' }
+  let(:stack_definition) { StackMaster::StackDefinition.new(base_dir: '/base_dir', region: region, stack_name: stack_name) }
+  let(:config) { double(find_stack: stack_definition) }
```
viraptor, line 5: instance_double?

lib/stack_master/commands/nag.rb
viraptor: Also, cfn_nag claims they support yaml: https://github.com/stelligent/cfn_nag/blob/master/lib/cfn-nag/cfn_nag.rb#L76

viraptor:
Looks good. I like the extra linting tools growing around cfn :)

lib/stack_master/commands/nag.rb
simpsora: You're right!  Tested and works; fixed in d3789e4.

simpsora:
I'm hoping we can build up some rules of our own that fit how we use CF as well.

lib/stack_master/commands/nag.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Commands
+    class Nag
+      include Command
+      include Commander::UI
+
+      def perform
+        Tempfile.open(['stack', "___#{stack_definition.stack_name}.#{proposed_stack.template_format}"]) do |f|
+          f.write(proposed_stack.template_body)
+          f.flush
+          system('cfn_nag', f.path)
```
patrobinson, line 11: How do we pass the exit code back out? When running in CI we'd want it to exit non-zero if the checks fail.

simpsora: Good catch @patrobinson 🤦 

Fixed in 46e5db7, including tests (which are a little weird, happy to hear other mechanisms for testing/mocking exit status).

I will note that I based my implementation on the `lint` command, which has no such behavior.  Probably worth a PR to fix that as well.

simpsora:
Comment from @stevehodgkiss:

> Apologies for not commenting on your PR for stackmaster for cfn-nag… I’m on the fence. Integrate cfn-nag vs stack_master compile us-east-1 stack | cfn-nag ... :thinking_face: I don’t have strong feelings either way :shrug:

I thought about that, and had these thoughts:
1.  `stack_master` is littered with `puts` statements going to stdout, so it's not following the unix philosophy and therefore is not possible to pipe it between tools.  This is fixable, if we desire to fix it.
2.  I originally looked into this issue because there's a coming PCI requirement to run SAST tools against our infra as code.  As it may/will become a requirement, integration seems like a worthwhile thing for us (though I acknowledge that this is an open-source tool).

Any strong feelings on this topic from anyone else?

lib/stack_master/commands/nag.rb
viraptor: This is unfortunately because the lint command exits with `0` only. :(

simpsora: Right, that makes sense.  I wonder if it would be reasonable to perform a string match on the output (assuming there's some sentinel value we could use to identify failure/success) and use that to exit properly? 

In any case, that's work for a future PR.

simpsora:
Stan has another SM PR in the works.  What's the standard practice for two new features ready at once?  Release a single new version with both features or do back-to-back separate releases?

patrobinson:
I'm ok with a single release with both features

patrobinson:
You can leave it in the Changelog as "unreleased" and then Stan can move it into that specific version

PR #343, opened 2020-07-07
Suggest other regions

viraptor:
When trying to apply a stack in a region where it's not defined we could do
better than just print the error.
List regions where the stack by that name is defined.

### DIFF
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -151,8 +151,8 @@ Feature: Apply command
     And the exit status should be 1
 
   Scenario: Run apply with stack in wrong region
-    When I run `stack_master apply foo myapp_web`
-    Then the output should contain "Stack name myapp_web exists in regions: us-east-1"
+    When I run `stack_master apply foo myapp-web`
+    Then the output should contain "Stack name myapp-web exists in regions: us-east-1"
     And the exit status should be 1
 
   Scenario: Create stack with --changed
--- a/features/apply.feature
+++ b/features/apply.feature
@@ -150,6 +150,11 @@ Feature: Apply command
     Then the output should contain "Could not find stack definition bar in region foo"
     And the exit status should be 1
 
+  Scenario: Run apply with stack in wrong region
+    When I run `stack_master apply foo myapp_web`
+    Then the output should contain "Stack name myapp_web exists in regions: us-east-1"
+    And the exit status should be 1
+
   Scenario: Create stack with --changed
     Given I stub the following stack events:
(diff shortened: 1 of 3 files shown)


lib/stack_master/cli.rb
```diff
@@ -270,6 +271,13 @@ def execute_stacks_command(command, args, options)
       @kernel.exit false unless success
     end
 
+    def show_other_region_candidates(stack_name)
+      candidates = config.filter(stack_name=stack_name)
+      return if candidates.empty?
+
+      StackMaster.stdout.puts "Stack name #{stack_name} exists in regions: #{candidates.map(:&region).join(', ')}"
```
simpsora: I wonder if this should be more clearly worded as an error case, with a helpful suggestion.  e.g.:

```suggestion
      StackMaster.stdout.puts "Stack name #{stack_name} not found in region #{region}, but was found in regions: #{candidates.map(:&region).join(', ')}"
```

viraptor: This is already a part of the previous line: https://github.com/envato/stack_master/pull/343/files#diff-f83645dd9bf35512a532426db4fed419R256

simpsora: Oh -- so it does.

simpsora:
Have merged https://github.com/envato/stack_master/pull/342, so feel free to merge that in here, update the docs, and we can release both at once.

features/apply.feature
```diff
@@ -150,6 +150,11 @@ Feature: Apply command
     Then the output should contain "Could not find stack definition bar in region foo"
     And the exit status should be 1
 
+  Scenario: Run apply with stack in wrong region
+    When I run `stack_master apply foo myapp-web`
+    Then the output should contain "Stack name myapp-web exists in regions: us-east-1"
+    And the exit status should be 1
+
```
simpsora, line 157: This is interesting, I opted for unit tests rather than features.  I wonder which is better 🤔

stevehodgkiss: This level of testing can also be done via rspec+aruba. I opted for that approach in the drift feature and I'm wondering if we should re-write all the cucumber features to use rspec instead. https://github.com/envato/stack_master/pull/339/files#diff-0845fa726c236bab4b7365489a3cd108R1

viraptor: In this case there was an existing cucumber test for the output of this command, so it was the easiest solution.

stevehodgkiss: Yep, 👍. Sorry, I wasn't suggesting that this change needed to move to rspec.

PR #347, opened 2020-09-28
Support cross-region SSM parameter resolution

mipearson:
Had a case where I was about to copy a bunch of SSM keys to another region and then realised that there was a better way.

Struggled with trying to write tests for this: the existing tests, in their use of mocking the AWS api, will return the same response regardless of which region or SSM key is queried. So this is technically "covered" - but not explicitly tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,9 @@ you will likely want to set the parameter to NoEcho in your template.
 ```yaml
 db_password:
   parameter_store: ssm_parameter_name
+saas_api_key:
+  # Parameter in a different region in the same account
+  parameter_store: us-east-1:ssm_parameter_name
 ```
 
 ### 1Password Lookup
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -307,6 +307,9 @@ you will likely want to set the parameter to NoEcho in your template.
 ```yaml
 db_password:
   parameter_store: ssm_parameter_name
+saas_api_key:
+  # Parameter in a different region in the same account
+  parameter_store: us-east-1:ssm_parameter_name
```
orien, line 312: An alternative interface for consideration:

```yaml
saas_api_key:
  parameter_store:
    name: ssm_parameter_name
    region: us-east-1
```

mipearson: Yeah, thought about it, but wanted to keep it consistent with how we do stack output cross-region things.

stevehodgkiss (approved):

orien (approved):

orien:
Don't forget to add an entry in the change log.

PR #348, opened 2020-10-02
Empty strings are valid

patrobinson:
I'm trying to write a dynamic that takes a compile time parameter and if it's set adds a statement to a policy.

But this code doesn't let me, apparently empty strings are not valid. I don't think that's true, empty strings are perfectly ok! Especially as we can't set nil values either.


stevehodgkiss (approved):
💯 I think `nil` should be allowed also

patrobinson:
Covered in this PR https://github.com/envato/stack_master/pull/199

patrobinson:
Haha, I basically reimplemented #199

PR #358, opened 2021-02-02
Make the compile time parameter error clearer

viraptor:
Print out the expectation that the parameter should be under a specific key.

### DIFF
--- a/lib/stack_master/sparkle_formation/compile_time/empty_validator.rb
+++ b/lib/stack_master/sparkle_formation/compile_time/empty_validator.rb
@@ -23,7 +23,7 @@ module StackMaster
         end
 
         def create_error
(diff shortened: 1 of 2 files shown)


lib/stack_master/sparkle_formation/compile_time/empty_validator.rb
```diff
@@ -23,7 +23,7 @@ def has_invalid_values?
         end
 
         def create_error
-          "#{@name} cannot contain empty parameters:#{@parameter.inspect}"
+          "#{@name} is not present inside the 'compile_time_parameters' entry, or has an empty value; found: #{@parameter.inspect}"
```
orien, line 26: I think we should inform that it's a parameter that's missing. Can sort of infer it from the `compile_time_parameters` name, but better to be explicit.

orien (approved):

Issue #362, opened 2021-10-04
Unclear error when missing permissions to validate allowed account


liamdawson:
When running `stack_master`, and the current account's ID isn't in the (non-empty) list of `allowed_account` values, it attempts to fetch account aliases to check if these match. If the current principal isn't permitted to `iam:ListAccountAliases`, this results in the following error:

```
$ stack_master validate ap-southeast-2
Executing validate on stack-name in ap-southeast-2
error: Failed to retrieve account aliases. Missing required IAM permission: iam:ListAccountAliases. Use --trace to view backtrace
```

It becomes a bit clearer if you use `--trace`:

```
$ stack_master validate ap-southeast-2 --trace
...
         4: from .../stack_master/lib/stack_master/cli.rb:294:in `execute_if_allowed_account'
         3: from .../stack_master/lib/stack_master/cli.rb:305:in `running_in_allowed_account?'
         2: from .../stack_master/lib/stack_master/identity.rb:10:in `running_in_account?'
         1: from .../stack_master/lib/stack_master/identity.rb:45:in `contains_account_alias?'
.../stack_master/lib/stack_master/identity.rb:22:in `account_aliases': Failed to retrieve account aliases. Missing required IAM permission: iam:ListAccountAliases (StackMaster::Identity::MissingIamPermissionsError)
```

Ideally, I think the error message returned to the user should make it immediately obvious why `stack_master` attempted to use that permission.

PR #363, opened 2021-10-04
Improve allowed account handling

liamdawson:
Per #362, there's some opportunity for improvement around allowed account logic. This PR introduces two changes:

1. Return a more contextual error message when  `Identity#account_aliases` raises an IAM permissions error in `Identity#running_in_account?` — previously, users had to examine the stack trace to work out which part of the application logic triggered the error.

    Note that the previous error message (identifying the missing IAM permission) is no longer shown, but can be seen with `--trace`, as the `MissingIamPermissionsError` is recorded as the `Error#cause` of the `AllowedAccountAliasesError`.
2. Avoid checking the current account's aliases against the `allowed_accounts` list if all values are 12-digit numbers (i.e. AWS account IDs) — considering the way AWS account aliases are used, it should not be possible to use a 12-digit number as an account alias

Some tests had to be updated when making this change, as they used AWS account IDs of an incorrect length.

### DIFF
--- a/features/apply_with_allowed_accounts.feature
+++ b/features/apply_with_allowed_accounts.feature
@@ -5,14 +5,14 @@ Feature: Apply command with allowed accounts
       """
       stack_defaults:
         allowed_accounts:
-          - '11111111'
+          - '111111111111'
       stacks:
         us_east_1:
           myapp_vpc:
             template: myapp.rb
           myapp_db:
             template: myapp.rb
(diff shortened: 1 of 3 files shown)


orien (approved):
👍

petervandoros (approved):

patrobinson:
@liamdawson changes are pretty infrequent so I think we'd be better bumping this with a minor version change.

PR #365, opened 2021-11-17
Exit with status code 1 if any command fails

craigforster:
Today, when a command is executed across multiple stacks the status of each stack isn't checked. This means if one returns an error, the command continues processing across the other stacks.

This PR updates `execute_stacks_command ` exit at the first error. 

This allows CI processes to run `validate` (or other commands) across a range of stacks and be confident that an errors result in a failed build.

### DIFF
--- a/lib/stack_master/cli.rb
+++ b/lib/stack_master/cli.rb
@@ -274,9 +274,9 @@ module StackMaster
         stack_definitions.each do |stack_definition|
           StackMaster.cloud_formation_driver.set_region(stack_definition.region)
           StackMaster.stdout.puts "Executing #{command.command_name} on #{stack_definition.stack_name} in #{stack_definition.region}"
-          success = success && execute_if_allowed_account(stack_definition.allowed_accounts) do
+          success = execute_if_allowed_account(stack_definition.allowed_accounts) do
             command.perform(config, stack_definition, options).success?
-          end
+          end && success
         end
       end
       @kernel.exit false unless success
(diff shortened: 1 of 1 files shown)


patrobinson:
Hi @craigforster 

Thanks for the PR.

It seems our behaviour is a bit inconsistent. I don't think we want to fail immediately if there's an invalid stack, however I do expect that we return with an exit code of something other than 0 which doesn't seem to always be the case. However it seems as though in other scenarios we do fail fast when I would expect us not to.

Scenario 1: Unable to resolve parameter
Outcome: Fail fast, exit code 1

Scenario 2: Empty/blank parameter
Outcome: No failure, exit code 0

This puts us in an unfortunate position of choosing to keep existing inconsistent behaviour or include a breaking change and have to bump the major version. I'm inclined to keep inconsistent behaviour for now, but at least fix the exit code so it's what we expect.

It'd be good to have a failing test to replicate this behaviour as well as the behaviour for the first scenario, so we ensure we avoid potentially breaking changes.

craigforster:
Hi @patrobinson 

I'll do my best to get up to speed enough on the code to add some tests -- fair warning, I'm not particularly experienced with Ruby.  Is there a particular spec you'd like me to add the test or tests to?

craigforster:
Another thought I had was to preserve the default behavior by default, and add a global option to enable `fail-fast` or the like.  This would preserve backwards compatibility and avoid a major version bump.

craigforster:
I'm going to close this PR, iterate locally, and then open a fresh one when I have a solution I'm happy with.  Any pointers to existing tests would be definitely appreciated.

patrobinson:
@craigforster having a fail fast option would help retain the same behaviour but give you what you want so I'm in favour of it.

I've written some failing tests on this branch https://github.com/envato/stack_master/tree/validate-exit-code. Feel free to cherry-pick them into your branch

craigforster:
Amazing, thanks!

PR #371, opened 2022-12-27
AWS SDK: pass an options hash instead of keyword arguments

orien:
The AWS SDK accepts an options hash as arguments, not keyword arguments. We've been relying on Ruby to automatically convert keyword arguments into an options hash. This implicit conversion is frowned upon with Ruby 3 and becomes a problem in Ruby 3.2.

 Let's avoid this and explicitly pass an options hash.


stergiom (approved):
👏 Nice. Sorry, late to review (missed this PR somehow).

PR #374, opened 2023-07-28
Use sorted_set gem instead of relying on it existing in ruby's `set` library

petervandoros:
Resolves the following error:

```
StackMaster::TemplateCompiler::TemplateCompilationFailed Failed to compile a_stack_template.rb
 Caused by: RuntimeError The `SortedSet` class has been extracted from the `set` library. You must use the `sorted_set` gem or other alternatives.
```

In ruby 3.1, the SortedSet class was removed from the `set` library.

The sparkle_formation gem has resolved this issue, but has yet to release a new version of the gem.

This commit works around the issue by ensuring the SortedSet class in the `sorted_set` gem is used by the sparkle_formation gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,16 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ## [Unreleased]
 
-[Unreleased]: https://github.com/envato/stack_master/compare/v2.13.3...HEAD
+[Unreleased]: https://github.com/envato/stack_master/compare/v2.13.4...HEAD
+
+## [2.13.4] - 2023-07-31
+
+### Fixed
+
+- Resolve error caused by `SortedSet` class being removed from the `set` library ([#374]).
+
+[2.13.3]: https://github.com/envato/stack_master/compare/v2.13.3...v2.13.4
+[#374]: https://github.com/envato/stack_master/pull/374
 
 ## [2.13.3] - 2023-02-01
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,16 @@ The format is based on [Keep a Changelog], and this project adheres to
...
 ## [Unreleased]
 
-[Unreleased]: https://github.com/envato/stack_master/compare/v2.13.3...HEAD
+[Unreleased]: https://github.com/envato/stack_master/compare/v2.13.4...HEAD
+
+## [2.13.4] - 2023-07-31
+
+### Fixed
+
+- Resolve error caused by `SortedSet` class being removed from the `set` library ([#374]).
```
orien, line 19: This is only a problem for those projects using SparkleFormation.

```suggestion
- Resolve SparkleFormation template error caused by `SortedSet` class being removed from the `set` library in Ruby 3 ([#374]).
```

orien (approved):
I hesitated to do this because the problem is not in StackMaster, and only impacts projects using SparkleFormation.

But this is the pragmatic approach. And it's the solution we'll eventually get from SparkleFormation anyway.

CHANGELOG.md
petervandoros: Thanks, fixed in [254eb03](https://github.com/envato/stack_master/pull/374/commits/254eb03871efc652a46fb542f15514c0e76a8b67).

runlevel5 (approved):

PR #380, opened 2024-03-07
Excplicit error on empty stack defaults

viraptor:
Prevent an exception when then stack_defaults are empty.


orien (approved):
I wonder if we can  define a schema and use a general-purpose validation tool to do this work.

viraptor:
That sounds way too reasonable... Maybe the next time I run into some failure.

PR #386, opened 2024-07-31
Support non-camel-case stack outputs

redterror:
This adds support for non-camel-case stack output keys.  In my specific case I had to handle a key like `fooBarBaz` set by another application.


orien (approved):
Thank you.

orien:
Released in v2.16.0.

PR #390, opened 2025-07-10
Add SSO group id lookup parameter resolver

andrewjhumphrey:
Remove the need to hardcode magic AWS IIC/SSO group IDs that are required when creating AWS::SSO::Assignment resources

Note:  This _requires_ adding a new stack_master.yml toplevel attribute: sso_identity_store_id which is required to do the group lookup.  Rather than specifying that in each individual parameter lookup, do it once at the top level.

An example of what this looks like:

## Before:
Parameters file:

```yaml
SSOArn:
  stack_output: sso-permission-sets-standard/SSOArn
  role: service-role/rolename
  account: '123456789012'
SecurityAdminPermissionSetArn:
  stack_output: sso-permission-sets-standard/SecurityAdminArn
  role: service-role/rolename
  account: '123456789012'
GroupId: "b12345c6-7890-12e3-45b6-b7f89dded0aa"   <---- Needs to be manually fetched from the console/cli
```

## After:

Parameters file:

```yaml
SSOArn:
  stack_output: sso-permission-sets-standard/SSOArn
  role: service-role/rolename
  account: '123456789012'
SecurityAdminPermissionSetArn:
  stack_output: sso-permission-sets-standard/SecurityAdminArn
  role: service-role/rolename
  account: '123456789012'
GroupId:
  sso_group_id: 'us-east-1:d-123456df7/Okta-App-AWS-Group-name'
  role: service-role/rolename
  account: '123456789012'
```


README.md
```diff
@@ -416,6 +416,24 @@ ssh_sg:
...
 
+### AWS IIC/SSO Group IDs
+
+Looks up AWS Identity Center group name in the configured Identity Store and returns the ID suitable for use in AWS IIC assignments.
+It is likely that account and role will need to be specified to do the lookup.
+
+In stack_master.yml
+
+```yaml
+sso_identity_store_id: `d-12345678`
```
orien, line 427: I'm not convinced this needs to be a top-level attribute.

orien: There's prior art of embedding multiple attributes in the parameter value:

```yaml
bucket_name:
  # Output from a stack in a different region
  stack_output: us-east-1:init-bucket/bucket_name
```

Could we do the same here?

```yaml
GroupId:
  sso_group_id: 'd-1234567df8:Okta-App-AWS-Group-name'
  role: service-role/rolename
  account: '123456789012'
```

lib/stack_master/sso_group_id_finder.rb
```diff
@@ -0,0 +1,35 @@
+module StackMaster
+  class SsoGroupIdFinder
+    SSOGroupNotFound = Class.new(StandardError)
+    SSOIdentityStoreInvalid = Class.new(StandardError)
```
orien, line 4: Unused?

lib/stack_master/sso_group_id_finder.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def initialize(region)
+      @client = Aws::IdentityStore::Client.new({ region: region })
+    end
+
+    def find(reference, identity_store_id)
+      raise ArgumentError, 'SSO Group Name must be a non-empty string' unless reference.is_a?(String) && !reference.empty?
+
+      next_token = nil
+      all_sso_groups = []
```
orien, line 14: Unused?

lib/stack_master/sso_group_id_finder.rb
```diff
@@ -0,0 +1,35 @@
...
+            max_results: 50
+          })
+
+          matching_group = response.groups.find { |group| group.display_name == reference }
+          return matching_group.group_id if matching_group
+          break unless response.next_token
+          next_token = response.next_token
+        end
+      rescue Aws::IdentityStore::Errors::ServiceError => e
+          puts "Error calling ListGroups: #{e.message}"
```
orien, line 29: I'm not sure the benefit in rescuing this error. It seems it could cause an `SsoGroupNotFound` error, when the group actually exists in AWS. That'd be confusing.

README.md
andrewjhumphrey: It would be _super_ unusual to need to refer to 2 different identity stores so it felt like you'd just be copying the same thing over and over and over again.

lib/stack_master/sso_group_id_finder.rb
```diff
@@ -0,0 +1,35 @@
+module StackMaster
+  class SsoGroupIdFinder
```
orien, line 2: Is it worth writing a unit test for this class?

lib/stack_master/sso_group_id_finder.rb
andrewjhumphrey: Fixed in d0c6d62

lib/stack_master/sso_group_id_finder.rb
andrewjhumphrey: Fixed in d0c6d62

lib/stack_master/sso_group_id_finder.rb
andrewjhumphrey: I hadn't bothered as they kinda got exercised by the other tests, but given how easy it is to get ChatGPT to write unit tests there is little reason not too.

andrewjhumphrey: 6de910d

lib/stack_master/version.rb
```diff
@@ -1,3 +1,3 @@
 module StackMaster
-  VERSION = "2.16.0"
+  VERSION = "2.17.8"
```
orien, line 2: A typo perhaps.

spec/stack_master/sso_group_id_finder_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  let(:region) { 'us-east-1' }
+  let(:identity_store_id) { 'd-12345678' }
+  let(:group_name) { 'AdminGroup' }
+  let(:group_id) { 'abc-123-group-id' }
+
+  let(:aws_client) { instance_double(Aws::IdentityStore::Client) }
+
+  subject(:finder) do
+    # Ruby 3+ keyword args fix: make sure new accepts keyword args
+    allow(Aws::IdentityStore::Client).to receive(:new).with(hash_including(region: region)).and_return(aws_client)
```
orien, line 11: Can we put the stubbing in a `before` block?

orien: Actually, `Aws::IdentityStore::Client.new` doesn't take keyword args. Rather, it takes a `Hash`.

This should work:
```ruby
allow(Aws::IdentityStore::Client).to receive(:new).with({ region: region }).and_return(aws_client)
```

README.md
orien: The idea that the parameter is looked up by attributes outside of what is provided in the parameter file is surprising. Additionally, the configuration values are stored in separate locations, often in different files. This distance in configuration increases the potential for readers to be confused.

While nesting the identity store ID in the parameter value will increase duplication, the lookup will be explicit and more understandable for engineers.

spec/stack_master/sso_group_id_finder_spec.rb
andrewjhumphrey: I finally read the error message properly, so I reverted a bunch of unncessary change and updated the calls to pass hashes rather than keywords :smile:

andrewjhumphrey: [f51bee8](https://github.com/envato/stack_master/pull/390/commits/f51bee82b36543a6237f03bfbcb26fadac4f68c1)
[3bf111c](https://github.com/envato/stack_master/pull/390/commits/3bf111c0e2d5ca4dc9f31d135293876f22d9df01)

lib/stack_master/version.rb
andrewjhumphrey: I'll pop it back when folks are happy with this code, it just makes life easier for local testing (I can just do a bundle install to make sure I've got the right version)

README.md
andrewjhumphrey: I've converted across to this format as it made it easy to specify the region the directory store is in,  in our case it's the same as where we create the assignments, but that is not necessarily the case.

andrewjhumphrey: [f55d905](https://github.com/envato/stack_master/pull/390/commits/f55d90506d8ce6f4ae7494cb9d4991fb0cf2903c)
[f702341](https://github.com/envato/stack_master/pull/390/commits/f702341b34676c2bbe81ddfa40ef47edd88e8787)
[af6e6bd](https://github.com/envato/stack_master/pull/390/commits/af6e6bdda9546da80a9842e30963a5eb9b453b8f)

lib/stack_master/sso_group_id_finder.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def find(reference)
+      output_regex = %r{(?:(?<region>[^:]+):)?(?<identity_store_id>[^:/]+)/(?<group_name>.+)}
+
+      if !reference.is_a?(String) || !(match = output_regex.match(reference))
+          raise ArgumentError, 'Sso group lookup parameter must be in the form of [region:]identity-store-id/group_name'
+      end
+
+      region = match[:region] || StackMaster.cloud_formation_driver.region
+      client = Aws::IdentityStore::Client.new({ region: region })
```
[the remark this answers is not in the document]
andrewjhumphrey: Yes!  Yes it does!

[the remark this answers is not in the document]
orien: Yay!

lib/stack_master/version.rb
```diff
@@ -1,3 +1,3 @@
 module StackMaster
-  VERSION = "2.16.0"
+  VERSION = "2.17.0"
```
orien, line 2: Let's add an entry in the changelog also.

orien (approved):
Awesome!

Thanks for indulging my many suggestions.

runlevel5 (approved):
LGTM 👍

